PR #2, opened 2020-08-10
Feat/347 setup bdd

eoln:
- fix linting issues with long lines in license text
- setup BDD test runner

### DIFF
--- a/src/shared/inspect.ts
+++ b/src/shared/inspect.ts
@@ -1,11 +1,13 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 /*****
  License
  --------------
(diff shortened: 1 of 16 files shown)


src/shared/inspect.ts
```diff
@@ -31,11 +33,12 @@ export const defaults = {
   COLOR: true
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kleyow, line 36: question: is this likely to be needed often? can the ignore be added to the config file? or is this a case where a global ignore might prove detrimental?

kleyow:
I don't seem to have review access on this repo @lewisdaly 

lgtm!

PR #3, opened 2020-08-10
Feat/347 relative paths

eoln:
use `~` to shorten relative paths


jest.config.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
-
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
kleyow, line 2: aren't you disabling this here https://github.com/mojaloop/thirdparty-scheme-adapter/pull/3/files#diff-e4403a877d80de653400d88d85e4801aR60 so that these wont be needed?

eoln: thanks! fixed!

PR #6, opened 2020-08-12
Feat/347 dockerization

eoln:
- separate Dockerfile definitions for inbound and outbound service
- script commands to build and run in separation dockerized inbound and outbound service


Outbound.Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM node:12.16.1-alpine as builder
```
kleyow, line 1: SUPER nit: a bit more "structure" would be nice. newline between group of related steps. I'll leave the grouping to your discretion.

```
FROM node:12.16.1-alpine as builder
USER root
WORKDIR /opt/outbound-thirdparty-scheme-adapter

RUN apk add --no-cache -t build-dependencies git make gcc g++ python libtool autoconf automake \
    && cd $(npm root -g)/npm \
    && npm config set unsafe-perm true \
    && npm install -g node-gyp 

COPY package.json package-lock.json* /opt/outbound-thirdparty-scheme-adapter/
RUN npm ci

COPY . /opt/outbound-thirdparty-scheme-adapter

FROM node:12.16.1-alpine
WORKDIR /opt/outbound-thirdparty-scheme-adapter

# Create empty log file & link stdout to the application log file
RUN mkdir ./logs && touch ./logs/combined.log
RUN ln -sf /dev/stdout ./logs/combined.log

# Create a non-root user: outbound-thirdparty-scheme-adapter-user 
RUN adduser -D outbound-thirdparty-scheme-adapter-user 
USER outbound-thirdparty-scheme-adapter-user 
COPY --chown=outbound-thirdparty-scheme-adapter-user  --from=builder /opt/outbound-thirdparty-scheme-adapter .

EXPOSE 4005
CMD ["npm", "run", "start:outbound"]
```

kleyow:
Big fan of Outbound and Inbound being split so they can be run independently btw.

Outbound.Dockerfile
eoln: fixed

eoln:
thank you @kleyow !

PR #8, opened 2020-08-13
Feat/347 integration tests

eoln:
run integration test in CI/CD pipeline


test/integration/shared/get-metrics.test.ts
```diff
@@ -0,0 +1,26 @@
+import axios from 'axios'
```
kleyow, line 1: question: won't these get maybe a little cluttered containing both Inbound and Outbound tests?

kleyow: eh. nvm. It'll probably be fine.

eoln: I would like to keep the same structure for integration tests so the output on console log is much more consistent

PR #9, opened 2020-08-14
refactor: move handlers out from server folder

eoln:
let keep handlers separated from http server related code


test/unit/cli.test.ts
```diff
@@ -25,9 +25,9 @@
  --------------
  ******/
 
-import { Server } from '@hapi/hapi'
 import Config from '~/shared/config'
 import { Handlers } from '~/server'
+import { Server } from '@hapi/hapi'
```
akapur99, line 30: barring style, is there a specific reason for the shift?

eoln: I would like to keep in `/server` only the code related to server setup and `handlers` on the same level where `model` and `domain` is placed so we get more flat folder structure

eoln: @akapur99 barring style fixed

PR #12, opened 2020-08-18
feat/RedisConnection & KVS & PubSub

eoln:
partial migration of lib/cache.js from sdk-scheme-adapter


.eslintrc.js
```diff
@@ -35,8 +35,7 @@ module.exports = {
     'cucumber/no-arrow-functions': 2,
     'import/default': 'warn',
     'import/extensions': 'off',
-    'max-len': ["warn", { "code": 120 }],
-    'sort-imports': 'error'
```
kleyow, line 39: question: was `sort-imports` giving you trouble?

we could maybe use https://www.npmjs.com/package/eslint-plugin-simple-import-sort in the future. no action needed here.

eoln: sort-import and VSCode ESLint plugin can make you crazy ;-P

PR #14, opened 2020-08-20
feat: add env variables needed for MojaloopRequests

kleyow:
need a bit of guidance here Pawel, if this is the correct way to handle env variables we need.

nothing is finalized, really just a draft


src/shared/config.ts
```diff
@@ -101,6 +164,56 @@ export const ConvictConfig = Convict<ServiceConfig>({
...
+        mutualTLS: {
+          enabled: false
+        },
+        creds: {
+          ca: '',
+          cert: '',
+          key: ''
+        }
+      },
+      test: {
```
lewisdaly, line 206: What's the purpose of `test` here?

src/shared/config.ts
```diff
@@ -46,6 +60,55 @@ export interface ServiceConfig {
...
+    WSO2_AUTH: {
+      staticToken: string;
+      tokenEndpoint: string;
+      clientKey: string;
+      clientSecret: string;
+      refreshSeconds: number;
+    };
+    TLS: {
+      inbound: {
+        mutualTLS: {
```
lewisdaly, line 81: I wonder if we can use the power of typescript to better express this configuration.

Maybe something like (and this is rough so forgive me)

```ts

type TLSCredential = {
  ca: Array<Buffer>;
  cert: Array<Buffer>
  key: Array<Buffer>
}

type Option<T> =  T | null;

const mySetCred: Option<TLSCredential> = {
  ca: [],
  cert: [],
  key: [],
}
```

src/shared/config.ts
kleyow: I'm not exactly sure. I just replicated https://github.com/mojaloop/sdk-scheme-adapter/blob/master/src/config.js 
The tls object is passed as is to the MojaloopRequest constructor. I can investigate an answer for you tomorrow (for me).

lewisdaly: Ah ok - yeah let's try and keep changes to the `sdk-standard-components` as light as possible, so we probably want to keep this the same for now.

Same goes for my other comment about type improvements.

src/shared/config.ts
kleyow: Yes absolutely. I plan to break it down into smaller chunks.
Wanted to get feedback on the way I'm loading file data into env since it feels a bit weird...(I should've stated that)
Thanks for the feedback.

eoln: if possible, please use `interface` instead of `type`

src/shared/config.ts
kleyow: Ok. If we are putting improvements on the back burner this should be ready.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mojaloop/sdk-pisp-scheme-adapter",
-  "version": "11.1.1",
+  "version": "11.2.1",
```
eoln, line 3: don't bump manually the version, there is `npm run release` ;-)

eoln (changes requested):

package.json
kleyow: done.

eoln (approved):
lgtm

PR #15, opened 2020-08-21
feat: kvs integration tests

eoln:


test/integration/redis/kvs.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  })
+
+  it('should GET what was SET', async (): Promise<void> => {
+    const values = [
+      { a: 1, b: true, c: 'C', d: {} },
+      true,
+      123,
+      'the-string'
+    ]
+    const counter = 1
```
lewisdaly, line 64: I'm not sure I understand the counter here - should it be incremented? Or why not just do
```ts
key = `key-1`
```

eoln: @lewisdaly you are right! no need for this counter. Fixed

PR #16, opened 2020-08-25
Feat/348 migrate psm

eoln:
- add javascript-state-machine declaration to ambient.d.ts
- because of class creation and methods injections incompatibility in javas-state-machine introducing PersistentModel class in place of PersistentStateMachine


sridharvoruganti (approved):
lgtm

PR #19, opened 2020-08-26
feat: add inbound POST thirdpartyRequests/transaction handling

kleyow:
first pass. lmk any comments.


src/interface/api-inbound-template.yaml
```diff
@@ -85,4 +85,142 @@ paths:
           $ref: '../../node_modules/@mojaloop/api-snippets/v1.0/openapi3/responses/index.yaml#/503'
       operationId: HelloGet
       summary: example get
-      description: The HTTP request GET /hello is used to return some example json.
\ No newline at end of file
+      description: The HTTP request GET /hello is used to return some example json.
+  /thirdpartyRequests/transactions:
+    #parameters:
```
lewisdaly, line 90: Should these be commented out?

src/handlers/inbound/thirdpartyRequests/transactions.ts
```diff
@@ -0,0 +1,63 @@
...
+  // todo: this might need to be a state machine in the future
+  if (!verifyConsentId(transactionRequest.consentId) ||
+      !verifyPispId(request.headers['FSPIOP-Source']) ||
+      !verifySourceAccountId(transactionRequest.sourceAccountId) ||
+      !validateGrantedConsent(transactionRequest.consentId)) {
+    // todo: change to a thirdparty specific error code here
+    return h.response(Enums.FSPIOPErrorCodes.CLIENT_ERROR).code(Enum.Http.ReturnCodes.BADREQUEST.CODE)
+  }
+
+  forwardPostQuoteRequestToPayee(transactionRequest)
```
lewisdaly, line 56: So if I'm reading this correctly, this is kicking off the quote? So eventually this will be our entrypoint for the state machine - right?

test/unit/domain/thirdpartyRequests.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    resetUuid()
+  })
+
+  it('buildPayeeQuoteRequestFromTptRequest should construct a proper quote request from tptRequest', (): void => {
+    const expectMockQuoteRequest = expect(buildPayeeQuoteRequestFromTptRequest(
+      postThirdpartyRequestsTransactionRequest.payload
+    ))
+
+    expectMockQuoteRequest.toEqual(expect.objectContaining({
+      quoteId: '00000000-0000-1000-8000-000000000001',
```
lewisdaly, line 62: Nice, I like these deterministic uuids

test/unit/domain/thirdpartyRequests.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    }))
+  })
+
+  it('forwardPostQuoteRequestToPayee should forward a quote request and forward it to a payee', (): void => {
+    // build expected quote request
+    const quoteRequest = buildPayeeQuoteRequestFromTptRequest(
+      postThirdpartyRequestsTransactionRequest.payload
+    )
+
+    resetUuid()
```
lewisdaly, line 81: could this be in a `beforeEach`?

src/handlers/inbound/thirdpartyRequests/transactions.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Handles a DFSP inbound POST /thirdpartyRequests/transaction request
+ */
+async function post (_context: Context, request: Request, h: ResponseToolkit): Promise<ResponseObject> {
+  const transactionRequest = request.payload as ThirdpartyTransactionRequest
+
+  // todo: this might need to be a state machine in the future
+  if (!verifyConsentId(transactionRequest.consentId) ||
+      !verifyPispId(request.headers['FSPIOP-Source']) ||
+      !verifySourceAccountId(transactionRequest.sourceAccountId) ||
+      !validateGrantedConsent(transactionRequest.consentId)) {
```
lewisdaly, line 51: Happy to leave these for now, but I imagine they will be async calls

src/interface/api-inbound-template.yaml
kleyow: For now, yes. There is some header middleware/plugin that needs to be implemented.

src/handlers/inbound/thirdpartyRequests/transactions.ts
kleyow: Yes. From my understanding this should kick off the payee's `sdk-scheme-adapter` inbound POST /quotes handler. From there the `sdk-scheme-adapter`'s state machine should take over.

test/unit/domain/thirdpartyRequests.test.ts
kleyow: Thank James' work on the sdk-scheme-adapter. My first pass was much more bare.

test/unit/domain/thirdpartyRequests.test.ts
kleyow: it is in before each. 

`forwardPostQuoteRequestToPayee` also uses the mock uuid. So I need to reset it here to make sure it sends a quote with matching uuids.

src/handlers/inbound/thirdpartyRequests/transactions.ts
kleyow: yep. they'll be exactly that in the future.

src/handlers/inbound/thirdpartyRequests/transactions.ts
kleyow: At least that is how it is in my head...

src/domain/thirdpartyRequests/transactions.ts
```diff
@@ -0,0 +1,85 @@
...
+    tls: config.SHARED.TLS,
+    jwsSign: config.SHARED.JWS_SIGN,
+    jwsSigningKey: <Buffer> config.SHARED.JWS_SIGNING_KEY
+  })
+
+  // payee fspid should be defined due to a prior GET /parties call.
+  mojaloopRequests.postQuotes(quote, <string> quote.payee.partyIdInfo.fspId)
+}
+
+// todo: create dfsp outbound calls for these checks.
```
eoln, line 70: if there is a plan to make async calls let keep these function async also

src/handlers/inbound/thirdpartyRequests/transactions.ts
eoln: If we already know they should be async why leave technical debt here?

src/interface/api-inbound-template.yaml
eoln: let add TODO why we want to keep this commented code here

eoln (changes requested):

src/handlers/inbound/thirdpartyRequests/transactions.ts
kleyow: ok. made them async.

src/interface/api-inbound-template.yaml
kleyow: done.

eoln (approved):
lgtm

PR #21, opened 2020-09-01
Test/interface types

eoln:
temporary disable collecting coverage from interface/types.ts <- this file isn't used by source code anywhere -> it should be enabled back when used.


sridharvoruganti (approved):
lgtm

PR #33, opened 2020-09-18
feat(api): first pass at updated api design

lewisdaly:


lewisdaly:
Going to force merge this in so we at least have _some_ docs.

PR #34, opened 2020-09-25
Fix/349 api inbound examples

chbhernani:


test/unit/handlers/outbound-api.test.ts
```diff
@@ -230,6 +230,7 @@ describe('Outbound API routes', (): void => {
     })
   })
 
+  // jest.setTimeout(30000)
```
eoln, line 233: you need to set up the deferred call to inbound `PUT /authorization` using  setTimeout with 50ms delay and do the await call to `POST /authorization` at outbound and test the response body. You shouldn't touch `/thridparty....` endpoints.
you shouldn't use server.inject (this is dedicated for unit test only)

something like this draft:

```
it('post /authorizations', async () => {

        setTimeout(async () => {
            await axios(PUT http://<inbound_host>/authorization proper_put_payload)  
         }, 100)

        result = await axios(POST http://<outbound_host>/authorization proper_post_payload)

         ... assert result ....

})

```

eoln (changes requested):

test/unit/handlers/outbound-api.test.ts
chbhernani: I removed the changes and tests passed. No changes over this file anymore.

eoln (approved):
lgtm

PR #37, opened 2020-09-30
fix: 1617 signchallenge uri

eoln:
`/signchallenge` endpoint is already implemented in mojaloop-simulator https://github.com/mojaloop/mojaloop-simulator/blob/pisp/master/src/simulator/api.yaml#L219. 

there not flexible to have it hardcoded so I migrated this path to configuration


kleyow (approved):
lgtm

PR #40, opened 2020-10-07
feat: 1671 extend config to have all sdk standard components endpoints

eoln:
- also fix braking change for v12 in sdk-standard-components


lewisdaly:
Hmm tests are failing @eoln

src/shared/config.ts
```diff
@@ -77,48 +78,19 @@ export interface ServiceConfig {
...
     PEER_ENDPOINT: string
-    ALS_ENDPOINT: string
-    QUOTES_ENDPOINT: string
-    TRANSFERS_ENDPOINT: string
-    BULK_TRANSFERS_ENDPOINT: string
+    ALS_ENDPOINT?: string
+    QUOTES_ENDPOINT?: string
+    TRANSFERS_ENDPOINT?: string
+    BULK_TRANSFERS_ENDPOINT?: string
+    THIRDPARTY_REQUESTS_ENDPOINT?: string
```
sridharvoruganti, line 85: It seems to be forgot to add `THIRDPARTY_REQUESTS_ENDPOINT` and `TRANSACTION_REQUEST_ENDPOINT` in config files

kleyow (commented):
lgtm after sridhar's comment is addressed

src/shared/config.ts
eoln: there are optional parameters, if not specified PEER_ENDPOINT will be used

sridharvoruganti (approved):

PR #41, opened 2020-10-20
fix: quote request `note` field

lewisdaly:
- Addresses an issue where we created an quote with a empty `note` string. The sdk-scheme-adapter inbound didn't like this, but since note is not required, I simply removed it for now. In the future, we could add `note` to the `ThirdpartyTransactionRequest` body, if we wish to expose it to end users.

### DIFF
--- a/src/domain/thirdpartyRequests/transactions.ts
+++ b/src/domain/thirdpartyRequests/transactions.ts
@@ -33,7 +33,6 @@ export function buildPayeeQuoteRequestFromTptRequest (request: ThirdpartyTransac
   const quoteRequest = {
     quoteId: uuid(),
     transactionId: uuid(),
-    note: '',
     transactionRequestId: request.transactionRequestId,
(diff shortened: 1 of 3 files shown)


eoln (approved):
lgtm

sridharvoruganti (approved):

elnyry-sam-k:
@lewisdaly , @eoln , @sridharvoruganti - can one of you please log a bug for this against the sdk-scheme-adapter? having that recorded will help get it fixed later.

Thanks!

lewisdaly:
@elnyry-sam-k, I feel like an empty string _should_ be treated as a validation error. Although it isn't in the spec, so maybe we should consider it a bug.

elnyry-sam-k:
@lewisdaly , yes the minimum length is indeed '1' for Note.. so it is expected.. I was thinking if the field itself wasn't being accepted..

PR #50, opened 2021-03-04
docs: P2PTransaction sequence diagram

eoln:


docs/sequence/PISPTransactionApi.puml
```diff
@@ -27,40 +38,34 @@ rnote right of PISP #Light
...
-...User checks the party name, and confirms...
+...User checks the party name, and confirms transaction to continue...
 
 == Initate Transaction ==
 rnote right of PISP #Light
-""POST /thirdpartyTransactions/{id}/initiate""
+POST /thirdpartyTransactions/{id}/initiate
 {
   sourceAccountId: "dfspa.alice.1234",
   consentId: "555",
```
lewisdaly, line 71: We no longer need to provide the `consentId` here - that information is contained by the `sourceAccountId`

(see: https://raw.githubusercontent.com/mojaloop/pisp/master/docs/out/transfer/api_calls_detailed_2.svg for more information)

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
 end note
+Switch -> 3p_DFSP_IBS: POST /thirdpartyRequests/transactions
+activate 3p_DFSP_IBS
+3p_DFSP_IBS --> Switch: 202 Accepted
+rnote right of 3p_DFSP_IBS: state: transactionRequestReceived
+note over 3p_DFSP_IBS
+  **transactionRequestState** is **RECEIVED**
+end note
+deactivate Switch
+3p_DFSP_IBS -> Switch: GET /consents/{consentId}
```
lewisdaly, line 104: I'm not sure I understand this call here - the consent should always be known by the DFSP

lewisdaly: I suppose we might want to talk to the auth service here, but I'd say we can rely on the DFSP also knowing about the consent here.

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+  DFSP informs PISP that **transactionId** has been allocated\n
+end note
+rnote right of 3p_DFSP_IBS #Light
+PUT /thirdpartyRequests/transaction/{id}
+{
+  transactionId: '1234-1234-1234-1234',
+  transactionRequestState: 'RECEIVED'
+}
+end note
+3p_DFSP_IBS -> Switch: PUT /thirdpartyRequests/transaction/{id}
```
lewisdaly, line 133: Just a thought (that we haven't covered before) - there could be a potential race condition between the PISP receiving the `PUT /thirdpartyRequests/transaction/{id}` and the `POST /authorizations`

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+deactivate Switch
+3p_DFSP_IBS -> 3p_DFSP_IBS: Lookup **consentId, sourceAccountid and pispId**.\nVerify that they exists, and consent is granted with a valid credential
+3p_DFSP_IBS -> 3p_DFSP_IBS: Generate **transactionId**: '1234-1234-1234-1234'
+note over 3p_DFSP_IBS
+  DFSP informs PISP that **transactionId** has been allocated\n
+end note
+rnote right of 3p_DFSP_IBS #Light
+PUT /thirdpartyRequests/transaction/{id}
+{
+  transactionId: '1234-1234-1234-1234',
```
lewisdaly, line 129: Just to clarify: the `transactionId` here is _NOT_ the same as the `{id}` field in the path - that id is referring to the `transactionRequestId` - right?

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+SDK_DFSP_OBS -> 3p_DFSP_IBS: 200 QuotesInformation
+deactivate SDK_DFSP_OBS
+note over 3p_DFSP_IBS
+  DFSP has the QuotesInformation, they can now ask the PISP for authorization
+  **transactionRequestState** is **PENDING**
+end note
+rnote right of 3p_DFSP_IBS: state: pendingAuthorizationRequest
+3p_DFSP_IBS -> SDK_DFSP_OBS: SYNC POST /authorization
+activate SDK_DFSP_OBS
+SDK_DFSP_OBS -> Switch: POST /authorization
```
lewisdaly, line 164: note: in the latest API draft, `POST /authorizations` is now `POST /thirdpartyRequests/authorizations` - but let's leave this for now, since I don't want to have to touch the transaction-requests-service until we get an e2e transfer working.

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+rnote right of ST8: state: approvalReceived
 
 
-ST8 -> Switch: ""PUT /authorizations/{id}""
-Switch --> ST8: ""202 Accepted""
-rnote right of ST8: ""state: pendingTransaction""
+ST8 -> Switch: PUT /authorizations/{id}
+activate Switch
+Switch --> ST8: 202 Accepted
+rnote right of ST8: state: pendingTransaction
```
lewisdaly, line 210: could this be something like `pendingTransactionFinalization` or something? `pendingTransaction` seems a little vague to me.

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+deactivate SDK_DFSP_IBS
+deactivate SDK_DFSP_OBS
+note over 3p_DFSP_IBS
+  DFSP checks the authorization with the **AUTH-SERVICE**.
+end note
 
-note over Switch
-  1. DFSP checks the authorization with the auth-service, or their own auth-service.
-  2. DFSP is happy, so initiates the ""POST /transfer""
+3p_DFSP_IBS -> Switch: POST /thridpartyRequest/transactions/{transactionRequestId}/authorizations
```
lewisdaly, line 225: nit: typo

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+activate Switch
+Switch --> 3p_DFSP_IBS: 202 Accepted
+deactivate 3p_DFSP_IBS
+rnote right of 3p_DFSP_IBS: state: pendingAuthorizationVerification
+Switch -> AUTHS: POST /thridpartyRequest/transactions/{transactionRequestId}/authorizations
```
lewisdaly, line 230: nit: typo

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+AUTHS --> Switch: 202 Accepted
+deactivate Switch
+
+note over AUTHS
+  lookup Consent
+  ensure sourceAccountId matches with Consent
+  verify signed challenge with publicKey for Consent
+  AuthorizationInformation is verified and **ACCEPTED**
+end note
+AUTHS -> Switch: PUT /thridpartyRequest/transactions/{transactionRequestId}/authorizations
```
lewisdaly, line 241: nit: typo

docs/sequence/PISPTransactionApi.puml
```diff
@@ -71,78 +76,214 @@ rnote right of PISP #Light
...
+activate Switch
+Switch --> AUTHS: 202 Accepted
+deactivate AUTHS
+Switch -> 3p_DFSP_IBS: PUT /thridpartyRequest/transactions/{transactionRequestId}/authorizations
```
lewisdaly, line 245: nit: typo

lewisdaly (commented):
Looks great @eoln. Just a few things that we might want to discuss, and other minor improvements.

I love how we have simplified the interactions with sdk-scheme-adapter - great work!

docs/sequence/PISPTransactionApi.puml
eoln: @lewisdaly if we drop `consentId` from this payload (and from `POST /thirdpartyRequests/transactions` send to DFSP), than DFSP needs to resolve `sourceAccountId (accountId)` to get `consentId`. DFSP needs to use a `Map[accountId] -> consentId` for this. 

I've checked our Auth-Service and I see that for given `consentId` we are storing `sourceAccountId` on the `scopes` list - so DFSP has all data needed for this mapping. But there is no interface exposed to make such a query on Auth-Service... 

I find out that `Map[accountId] -> consentId` mapping and quering isn't trival. We can have multiple `consentId` registered for the same `accountId` 

Because from the specification of `/accounts/{UserId}` I see that we are having a list of available accounts for the user for the given DFSP. This list is used during linking a device with an account and represented in the system by Consent and its id. And one of these accounts will be used as `sourceAccountId` in our payload.

In the case when the user has multiple devices (old & new phone for example) we would have registered more than one Consent for the same AccountId. 

So the map is now `Map[accountId] -> [consentIdOldPhone, consentIdNewPhone]` 

It brings us the problem - How DFSP will recognize which `consentId` from the mapping should it use to validate POST /thirdpartyRequests/transactions ?

docs/sequence/PISPTransactionApi.puml
eoln: from which part of the context for this workflow DFSP will derive about `Consent`? Currently we are getting `consentId` in the payload: https://github.com/mojaloop/thirdparty-api-adapter/blob/master/src/interface/api-template.yaml#L806-L863, so DFSP needs to get details about - in our scenario DFSP retrieves it from AUTH-SERVICE

docs/sequence/PISPTransactionApi.puml
eoln: In real life there shouldn't be any race condition, the DFSPTransactionModel workflow is blocked on SYNC POST /quotes on sdk-scheme-adapter before calling SYNC  POST /authorization. 

But we can mitigate this hazardous risk by proper state two-step transition paths in PISPTransactionModel. If we specify that transitions: `transactionRequestReceived` and `authorizationReqeustReceived` can be done independently and the final state `transactionInitiationSuccess` can be only reached when both are fulfilled

excerpt from the state machine definition
```
transitions : [
....
{ name: 'transactionRequestReceived', from: 'pendingTransactionInitiation', to: 'transactionRequestReceived' },
{ name: 'transactionRequestReceived', from: 'authorizationRequestReceived', to: 'transactionInitiationSuccess' },
{ name: 'authorizationRequestReceived', from: 'pendingTransactionInitiation', to: 'autorizationsRequestReceived' },
{ name: 'authorizationRequestReceived', from: 'transactionRequestReceived', to: 'transactionInitiationSuccess' }
....
]
```

docs/sequence/PISPTransactionApi.puml
eoln: yes, you are right I will fix it properly

eoln:
@lewisdaly & @millerabel I updated the diagram, more details visualized.
I point out two workflow models PISPTransactionModel to handle PISP side, and complementary DFSPTransactionModel to implement transfer flow.

docs/sequence/PISPTransactionApi.puml
eoln: I renamed it to `pendingVerifyAuthorization`

docs/sequence/PISPTransactionApi.puml
lewisdaly: Thanks for this careful exposition, @eoln.

>than DFSP needs to resolve byself sourceAccountId (accountId) to get consentId
We are also dropping the `sourceAccountId`. In the latest API Draft, Michael removed `sourceAccountId` and `consentId`, in favour of that information being contained in the payer `PartyIdInfo`. When a link is created, for each account that is linked, a `THIRD_PARTY_LINK` identifier type will be registered at the ALS, and the DFSP will be able to infer (1) the source of funds account and (2) the `consentId` based on the `payer`field alone:

Screen Shot 2021-03-08 at 8 35 34 pm

ref: https://raw.githubusercontent.com/mojaloop/pisp/master/docs/out/transfer/api_calls_detailed_2.svg

An `accountId` should be unique for a User, Device, PISP and DFSP, so it shouldn't be reused by a DFSP for a subsequent linking request with a different device.

As a part of the linking process, we expect the DFSP to keep a record of the `Map[accountId] -> consentId` mapping... perhaps for your implemention now this will live in the `mojaloop-simulator`? Or could we reasonably expect the `thirdparty-scheme-adapter` to remember this type of information?

docs/sequence/PISPTransactionApi.puml
lewisdaly: I don't think that's right. I think the DFSP needs to keep a record of the `consentId`, as well as (optionally) storing it with the Auth-Service. Let's discuss more after the standup.

docs/sequence/PISPTransactionApi.puml
lewisdaly: Awesome, yeah I like the sound of this, if it doesn't add too much complexity to implement.

lewisdaly:
Thanks @eoln - I expect we will need a little more discussion on the `consentId` stuff.

eoln:
**Question:** @lewisdaly @millerabel @MichaelJBRichards :  maybe we should notify PISP
  that Authorization received is OK
  and DFSp will initiate Transfer

eoln:
As we discussed yesterday with @lewisdaly I dropped AUTH-SERVICE in favor of delegating the AUTH-SERVICE (which is an optional service) related logic to DFSP backend which will be implemented by TTK

mjbrichards:
> **Question:** @lewisdaly @millerabel @MichaelJBRichards : maybe we should notify PISP
> that Authorization received is OK
> let call `PATCH /thridpartyRequests/transactions/{transactionRequestId}` before DFSP will initiate Transfer

We should notify the PISP if the authorisation is rejected, but as far as I can see that isn't modelled in the sequence diagram at the moment. If the authorisation is accepted, I'm not sure that we need to send the PISP a message. It should be expecting a confirmation when the transfer succeeds. We don't expect to do this in the FSPIOP API (see table 49).

docs/sequence/PISPTransactionApi.puml
eoln: I've delegated the task to find out the Consent and verify it to DFSP backend.verifyConsent(pispId, payer)

docs/sequence/PISPTransactionApi.puml
```diff
@@ -27,125 +41,317 @@ rnote right of PISP #Light
...
+DFSP_TM -> DFSP_TM: SDKRequest.requestAuthorization(QuotesInformation)
+DFSP_TM -> DFSP_SDK_OUT: SYNC POST /authorization
+note over DFSP_TM
+  From now DFSPTransactionModel is blocked
+  on synchronous call to POST /authorization
+  until PISP initiate **Approve Transaction** phase.
+  Here timeout scenario should be handled properly
+end note
+activate DFSP_SDK_OUT
+DFSP_SDK_OUT -> Switch: POST /authorization
```
lewisdaly, line 201: nit: This should be `authorizations`

docs/sequence/PISPTransactionApi.puml
```diff
@@ -27,125 +41,317 @@ rnote right of PISP #Light
...
 
 
-ST8 -> Switch: ""PUT /authorizations/{id}""
-Switch --> ST8: ""202 Accepted""
-rnote right of ST8: ""state: pendingTransaction""
 
-note over Switch
-  1. DFSP checks the authorization with the auth-service, or their own auth-service.
-  2. DFSP is happy, so initiates the ""POST /transfer""
+note over PISP, DFSP_TM
```
lewisdaly, line 303: This is being cut off on the `.png` render - I think you need to give the puml renderer more memory (or switch to svg)

lewisdaly (changes requested):

docs/sequence/PISPTransactionApi.puml
```diff
@@ -27,125 +41,284 @@ rnote right of PISP #Light
...
+    -> (CASE-2) authorizationRequestReceived -> transactionRequestReceived
+  -> transactionInitiationSuccess (END)
+
+  CASE-1 is visualised here
+end note
 
-note over Switch
-  Switch kicks off the transaction - DFSP issues ""POST /authorizations""
+note over DFSP_TM
+  DFSP requires QuotesInformation\n
```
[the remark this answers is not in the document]
eoln: fixed

docs/sequence/PISPTransactionApi.puml
eoln: fixed

docs/sequence/PISPTransactionApi.puml
eoln: replaced png by svg

eoln:
I extended the diagram with a non-happy scenario for Authorization failure

lewisdaly (approved):

lewisdaly:
Thanks @eoln

PR #59, opened 2021-03-11
feat: 2060 separate backend requests and sdk requests

eoln:
Separate backend-requests to DFSPBackendReqeusts & PISPBackendRequest so now the code has a better structure


sridharvoruganti (approved):
lgtm

PR #62, opened 2021-03-16
fix: unresolved promise warning in authorization model

eoln:


test/unit/models/outbound/authorizations.model.test.ts
```diff
@@ -268,15 +268,15 @@ describe('OutboundAuthorizationsModel', () => {
...
       it('exceptions - Error', async () => {
-        const error = new Error()
+        const error = new Error('the-exception')
         mocked(modelConfig.requests.postAuthorizations).mockImplementationOnce(
           () => {
             throw error
           }
         )
-        const model = await create({ ...data, currentState: 'errored' }, modelConfig)
+        const model = await create({ ...data, currentState: 'start' }, modelConfig)
```
eoln, line 277: @sridharvoruganti currentState should be 'start' because we want to test exception thrown from postAuthorizations

sridharvoruganti (approved):

PR #63, opened 2021-03-16
feat: 2060 async2sync model

eoln:
migrated async2sync model from sdk-scheme-adapter.


kleyow (approved):
You make this look too easy Pawel...

PR #66, opened 2021-03-22
feat: handle outbound /consentRequests/{ID}/validate requests for OTP validation flow

kleyow:


kleyow:
Still need to do integration tests and maybe more error testing...but think this is a good point to get some extra eyes on this.

src/handlers/outbound/consentRequests/{ID}/validate.ts
```diff
@@ -0,0 +1,77 @@
...
+    consentRequestId,
+    h.getKVS(),
+    h.getLogger(),
+    h.getPubSub(),
+    h.getThirdpartyRequests()
+  )
+
+  const model = await create(data, config)
+
+  const result = (await model.run(args)) as unknown as OutboundOTPValidateResponse
```
eoln, line 69: if `create` has specified generic parameters properly, there shouldn't be necessary to do a type casting for `run` result
```suggestion
  const result: OutboundOTPValidateResponse = await model.run(args)
```

src/models/OTPValidate.model.ts
```diff
@@ -0,0 +1,99 @@
...
+import { PubSub, Message } from '../shared/pub-sub'
+import { ThirdpartyRequests, Logger as SDKLogger } from '@mojaloop/sdk-standard-components'
+import { OutboundOTPValidateResponse } from './OTPValidate.interface'
+
+export interface OTPValidateModelArgs {
+  consentRequestId: string,
+  fspId?: string,
+  consentRequest?: tpAPI.Schemas.ConsentRequestsIDPatchRequest
+}
+
```
eoln, line 43: I would define and export the model type:

```
export type OTPValidateModel = A2SModel<OTPValidateModelArgs, OutboundOTPValidateResponse>
```

src/models/OTPValidate.model.ts
```diff
@@ -0,0 +1,99 @@
...
+      throw new Error('TransfersModel args requires \'args.fspId\' to be nonempty string')
+    }
+  }
+
+  reformatMessage (message: Message) {
+    const messageObj = message as Record<string, unknown>
+    if (messageObj.errorInformation) {
+      return {
+        ...messageObj
+      } as OutboundOTPValidateResponse
```
eoln, line 92: `OutboundOTPValidateResponse` has mandatory property `consent` and optional `errorInformation`.  Is it valid to have `consent` undefined when `errorInformation` is present? I think no - it breaks this interface and as a result, the response from `run` method will also be broken. Maybe it is worth considering `consent` property to be conditional...

src/models/OTPValidate.model.ts
```diff
@@ -0,0 +1,99 @@
...
+  async requestAction (args: OTPValidateModelArgs): Promise<void> {
+    if (!args.fspId) {
+      throw new Error('OTPValidate args requires \'fspId\' to be nonempty string')
+    }
+
+    if (!(args.consentRequest && typeof (args.consentRequest) === 'object')) {
+      throw new Error('OTPValidate.requestAction args requires \'transfer\' to be specified')
+    }
+    this.thirdpartyRequests.patchConsentRequests(args.consentRequestId, args.consentRequest, args.fspId)
+    resolve()
```
eoln, line 75: let never use `path.resolve` when Promise<void> is expected. It is an antipattern and brings not-needed dependency and note expected side effects: https://nodejs.org/api/path.html#path_path_resolve_paths

When async function doesn't return any value it will be transformed into `Promise<void>` according to its signature - but `await` expression in the body should be present.

If you would like to be explicit then use promise factory: `return Promise.resolve()`

src/models/OTPValidate.model.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async requestAction (args: OTPValidateModelArgs): Promise<void> {
+    if (!args.fspId) {
+      throw new Error('OTPValidate args requires \'fspId\' to be nonempty string')
+    }
+
+    if (!(args.consentRequest && typeof (args.consentRequest) === 'object')) {
+      throw new Error('OTPValidate.requestAction args requires \'transfer\' to be specified')
+    }
+    this.thirdpartyRequests.patchConsentRequests(args.consentRequestId, args.consentRequest, args.fspId)
```
eoln, line 74: here we should await on patchConsentRequest to be resolved
```suggestion
    await this.thirdpartyRequests.patchConsentRequests(args.consentRequestId, args.consentRequest, args.fspId)
```

in A2SModel: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/models/a2s.model.ts#L143
we intentionally `await` on `requestAction` and continue workflow after it resolves. 

If we don't `await` here, then the `init` will be resolved before the `patchConsentRequests` API call will be made, what could potentially break the async workflow.

that is why the `requestAction` is declared as an `async function` that enforces using `await` expression or returning Promise - the workaround using `path.resolve` silences only the TS compiler.

src/models/inbound/consentRequests.model.ts
```diff
@@ -99,13 +100,11 @@ export class InboundConsentRequestsRequestModel {
       this.logger.push({ err }).error('Error in patchConsentRequest @ Inbound')
       const mojaloopError = this.reformatError(err)
```
eoln, line 101: where `reformatError` function is defined in `InboundConsentRequestsRequestModel` class?
@sridharvoruganti ^^^

eoln: there should be a unit test for consentRequests.model.ts: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/test/unit/models/inbound/consentRequests.model.test.ts
which covers unhappy scenarios and proper use of putConsentRequestsError

eoln (changes requested):

test/unit/models/inbound/consentRequests.model.test.ts
```diff
@@ -51,6 +53,7 @@ describe('InboundConsentRequestsRequestModel', () => {
     beforeEach(async () => {
       thirdpartyRequests = {
         postConsents: jest.fn(() => Promise.resolve()),
+        putConsentRequestsError: jest.fn(() => Promise.resolve()),
       } as unknown as ThirdpartyRequests
 
       dfspBackendRequests = {
```
eoln, line 59: `getScopes` is returning an Array as the root object, it is antipattern for JSON based APIs - there is a legacy security issue (any DFSP will start complaining when they have to implement it on their side):
- https://stackoverflow.com/questions/3503102/what-are-top-level-json-arrays-and-why-are-they-a-security-risk
- https://medium.com/@rm.dev/securing-json-apis-with-wrapper-objects-top-level-json-array-security-issue-1cf855468b8c

let use data property instead
```{ scopes: [....] }```

Another good point to avoid root Arrays is how easy it would be to extend response (if needed in the future): for Object root, you can simply add new optional properties without breaking the interface for old clients whereas using Array root you are locked down.

eoln (changes requested):

kleyow:
Going to break this up into several PR's.

PR #67, opened 2021-03-23
docs: add sequence diagrams for consent request OTP validation

kleyow:


eoln (changes requested):
let have svg generated files, and use them in docs - using puml server is unstable and often images aren't rendered at all.

kleyow:
done.

docs/sequence/DFSPOTPValidateModel.puml
```diff
@@ -0,0 +1,19 @@
+@startuml DFSPOTPValidateModel
```
eoln, line 1: I would suggest:
- introducing two final states: success / errored -> where on errored we call  putConsentRequestsError
- introduce transitions for happy flow & corresponding states:  
   - validateOTP (from=start, to=OTPisValid)
   - requestScopes (from=OTPisValid, to=scopesReceived)
   - storeConsent(from=scopesReceived, to=succeed)
   
  Then you can implement this workflow as an extension of PersistenModel class

eoln (changes requested):

eoln (approved):

PR #69, opened 2021-03-24
chore: introduce patch /consentRequests and update ttk spec

kleyow:
- Introduce outbound patch patch /consentRequests/{ID}/validate endpoint to outbound interface and ttk
- Update ttk references to v1_1

### DIFF
--- a/src/interface/api-outbound.yaml
+++ b/src/interface/api-outbound.yaml
@@ -292,6 +292,48 @@ paths:
           $ref: '#/components/responses/501'
         '503':
           $ref: '#/components/responses/503'
+  '/consentRequests/{ID}/validate':
+    patch:
+      operationId: OutboundConsentRequestsValidatePatch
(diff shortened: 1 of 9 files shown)


src/interface/api-outbound.yaml
```diff
@@ -478,6 +520,17 @@ components:
...
               currency: USD
+    ConsentRequestsValidateResponse:
+      description: Consent requests validate response
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              consent:
+                $ref: '#/components/schemas/ConsentsPostRequest'
```
eoln, line 531: a Request as a Response part - it doesn't fit to me ;-)

eoln (approved):

src/interface/api-outbound.yaml
kleyow: yea...we'll revisit this...

PR #72, opened 2021-03-25
refactor: rework inbound DFSP otp validate workflow

kleyow:


src/models/inbound/dfspOTPValidate.model.ts
```diff
@@ -0,0 +1,356 @@
...
+          await this.saveToKVS()
+          // workflow is finished
+          return
+
+        default:
+          this.logger.info('State machine in errored state')
+          return
+      }
+      this.logger.info(`Transfer model state machine transition completed in state: ${this.fsm.state}. Recursing to handle next transition.`)
+      this.run()
```
eoln, line 297: We are making here a recursive call that could be dangerous in javascript.
Recursion must be guarded. 
I can see that only two `cases` end with `break`, and the rest end with `return`. 
So there is a guard and its logic is spread across the switch cases - fine, but code layout could be improved:

I suggest replacing `break` with `return this.run()` so the whole case logic will be together in one block and there will be no need to remember that there is a recursion call after the switch statement or revising the switch logic when reader see the recursion call after the switch statement.

src/models/inbound/dfspOTPValidate.interface.ts
```diff
@@ -0,0 +1,76 @@
...
+  onSendConsent: Method
+}
+
+export interface DFSPOTPValidateModelConfig extends PersistentModelConfig {
+  pubSub: PubSub
+  thirdpartyRequests: ThirdpartyRequests
+  dfspBackendRequests: DFSPBackendRequests
+}
+
+// state machine does not return a response as an inbound handler state machine
```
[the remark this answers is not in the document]
eoln: definitely somewhere we should mention that this model is the passive inbound handler and the response shouldn't be generated synchronously from `run`

eoln (changes requested):

src/handlers/inbound/consentRequests/{ID}.ts
```diff
@@ -46,18 +47,26 @@ async function patch (_context: unknown, request: Request, h: StateResponseToolk
...
+    thirdpartyRequests: h.getThirdpartyRequests(),
   }
-  const model = new InboundConsentRequestsRequestModel(modelConfig)
+  const model = new DFSPOTPValidateModel(data, modelConfig)
 
   // don't await on promise to be resolved
   setImmediate(async () => {
-    const result = await model.postConsentsRequest(consentRequestsRequestId, sourceFspId, authToken)
-    modelConfig.logger.push({ result }).info('postConsentsRequest done')
+    await model.run()
```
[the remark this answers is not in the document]
kleyow: done. 

> how the client will know that the patch operation has been performed successfully?
via. a POST /consents request from the DFSP to the PISP

> if the error callback is called as a result of problems with handling the patch request how will it affect the flow?
I'm guessing the PISP will need to ask the user to retry an OTP on their end.

src/models/inbound/dfspOTPValidate.interface.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export enum DFSPOTPValidateModelState {
+  start = 'start',
+  validateOTP = 'validateOTP',
+  requestScopes = 'requestScopes',
+  sendConsent = 'sendConsent',
+  errored = 'errored'
+}
+
+export interface BackendValidateOTPResponse {
```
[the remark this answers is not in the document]
kleyow: There's no backend schema that generates the DTO's for this, or am I mistaken?

src/models/inbound/dfspOTPValidate.interface.ts
```diff
@@ -0,0 +1,76 @@
...
+export interface BackendGetScopesResponse {
+  scopes: tpAPI.Schemas.Scope[]
+}
+
+export interface DFSPOTPValidateStateMachine extends ControlledStateMachine {
+  validateOTP: Method
+  onValidateOTP: Method
+  requestScopes: Method
+  onRequestScopes: Method
+  sendConsent: Method
```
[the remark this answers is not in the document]
kleyow: done.

src/models/inbound/dfspOTPValidate.interface.ts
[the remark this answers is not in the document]
kleyow: sounds good. added a comment on the `DFSPOTPValidateModel` class.

src/models/inbound/dfspOTPValidate.model.ts
```diff
@@ -0,0 +1,356 @@
...
+      }
+    })
+    */
+
+  }
+
+  /**
+   * depending on current state of model returns proper result
+   */
+  getResponse (): void {
```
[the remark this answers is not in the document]
kleyow: yep copy pasta, sorry. removed

src/models/inbound/dfspOTPValidate.model.ts
kleyow: done.

src/handlers/inbound/consentRequests/{ID}.ts
```diff
@@ -66,7 +68,18 @@ async function patch (_context: unknown, request: Request, h: StateResponseToolk
 
   // don't await on promise to be resolved
   setImmediate(async () => {
-    await model.run()
+    try {
+      await model.run()
+    } catch (error) {
```
eoln, line 73: where are we inspecting or passing the `error` to the client?
-> maybe w should also dump it to the log

eoln (changes requested):

src/handlers/inbound/consentRequests/{ID}.ts
kleyow: What kind of error are you thinking? I was assuming if the state machine didn't run it would just be a server error. The other cases of errors are handled in each individual transistion.

src/models/inbound/dfspOTPValidate.interface.ts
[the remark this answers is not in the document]
eoln: aaach you are right - it is inbound interface -> so type should already be generated in api-snippets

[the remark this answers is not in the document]
eoln: aach - you are right!

src/handlers/inbound/consentRequests/{ID}.ts
eoln: in run method, in its global catch, we transit fsm to 'errored' state and then rethrow received exception so we got it here as an `error` parameter of catch. We should at least dump this error to log.

src/models/inbound/dfspOTPValidate.model.ts
eoln: hmm did you pushed changes? looks I am still seeing old version...

kleyow: yes I updated `break` to `return this.run()`

src/handlers/inbound/consentRequests/{ID}.ts
kleyow: ok. I forgot that I'm rethrowing the errors in trasistions. So I'll just log the error here and count on the transistions to send back PUT consentRequests/{ID}/error request.

src/models/inbound/dfspOTPValidate.model.ts
```diff
@@ -0,0 +1,347 @@
...
+
+        resolve()
+      } catch (error) {
+      }
+    })
+    */
+
+  }
+
+  reformatError (err: Error): Errors.MojaloopApiErrorObject {
```
lewisdaly, line 224: Should this be a util function implemented elsewhere? Is there any reason for it to be in this model implementation?

lewisdaly (approved):

src/models/inbound/dfspOTPValidate.model.ts
kleyow: Yes, it's used in a lot of places so it can be abstracted out. I've just put this here since moving it to a shared util folder would mean touching a lot of other files.

eoln (approved):

PR #75, opened 2021-03-29
feat: add model and handler of outbound pisp otp validate

kleyow:


src/models/outbound/pispOTPValidate.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+  authToken: string
+  toParticipantId: string
+}
+
+export interface OutboundOTPValidateConsentResponse {
+  consent: tpAPI.Schemas.ConsentsPostRequest
+  currentState: PISPOTPValidateModelState
+}
+
+export interface OutboundOTPValidateErrorResponse {
```
eoln, line 72: same here - could we use DTO interface generated from OpenAPI?

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -0,0 +1,236 @@
...
+import { ThirdpartyRequests } from '@mojaloop/sdk-standard-components';
+import { thirdparty as tpAPI, v1_1 as fspiopAPI } from '@mojaloop/api-snippets';
+import inspect from '~/shared/inspect'
+import {
+  PISPOTPValidateData,
+  PISPOTPValidateStateMachine,
+  PISPOTPValidateModelConfig
+} from './pispOTPValidate.interface';
+import { Message } from '../../shared/pub-sub';
+import { OutboundOTPValidateConsentResponse, OutboundOTPValidateErrorResponse } from './pispOTPValidate.interface';
```
eoln, line 40: let group all imports from `./pispOTPValidate.interface` in one block

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -0,0 +1,236 @@
...
+    const { consentRequestsRequestId, authToken, toDFSPParticipantId } = this.data
+
+    const channel = PISPOTPValidateModel.notificationChannel(
+      consentRequestsRequestId
+    )
+
+    this.logger.push({ channel }).info('onValidateOTP - subscribe to channel')
+
+    // eslint-disable-next-line no-async-promise-executor
+    return new Promise(async (resolve, reject) => {
```
eoln, line 94: let use here `deferredJob`

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -0,0 +1,236 @@
...
+        } as OutboundOTPValidateErrorResponse
+      default:
+    }
+  }
+
+  // utility function to check if an error after a transistion which
+  // pub/subs for a response that can return a mojaloop error
+  async checkModelDataForErrorInformation(): Promise<void> {
+    if (this.data.errorInformation) {
+      await this.fsm.error()
```
eoln, line 149: `fsm.error` accepts as parameter the `error`

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -0,0 +1,236 @@
...
+
+        // avoid circular ref between PISPOTPValidateState.lastError and err
+        err.PISPOTPValidateState = { ...this.data }
+      }
+      throw err
+    }
+  }
+}
+
+export async function existsInKVS (config: PISPOTPValidateModelConfig): Promise<boolean> {
```
eoln, line 198: do we really need this method?

test/integration/outbound/pisp-otp-validate.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  describe('/consentRequests/{ID}/validate: requestAction->OTPIsValid', (): void => {
+    it('OTPValidateState should be OTPIsValid', async (): Promise<void> => {
+      const consentRequestsRequest = {
+        toParticipantId: 'dfspa',
+        authToken: '123456'
+      }
+      const expectedResponse = {
+        "consentId": "8e34f91d-d078-4077-8263-2c047876fcf6",
+        "consentRequestId": "ab049ac1-b936-4327-ab76-3e66e608beb8",
```
eoln, line 65:
```suggestion
        "consentRequestId": consentRequestsRequestId,
```
let reuse defined const

eoln (changes requested):

src/handlers/outbound/consentRequests/{ID}/validate.ts
```diff
@@ -58,6 +58,7 @@ async function patch (_context: any, request: Request, h: StateResponseToolkit):
     key: consentRequestsRequestId,
     logger: h.getLogger(),
     thirdpartyRequests: h.getThirdpartyRequests(),
+    requestProcessingTimeoutSeconds: 3
```
eoln, line 61: shouldn't we take this value from config?

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -90,37 +91,35 @@ export class PISPOTPValidateModel
...
     this.logger.push({ channel }).info('onValidateOTP - subscribe to channel')
 
-    // eslint-disable-next-line no-async-promise-executor
-    return new Promise(async (resolve, reject) => {
-      let subId = 0
+    return deferredJob(this.pubSub, channel)
+    .init(async (channel) => {
+      const res = await this.thirdpartyRequests.patchConsentRequests(
+        consentRequestsRequestId,
+        { authToken: authToken }  as tpAPI.Schemas.ConsentRequestsIDPatchRequest,
```
eoln, line 98:
```suggestion
        { authToken }
```
the type casting to as `tpAPI.Schemas.ConsentRequestsIDPatchRequest` 
shouldn't be necessary here.

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -90,37 +91,35 @@ export class PISPOTPValidateModel
...
+        toParticipantId
+      )
+
+      this.logger.push({ res, channel })
+        .log('ThirdpartyRequests.patchConsentRequests request call sent to peer, listening on response')
+      return Promise.resolve()
```
eoln, line 104: we can drop this line, it is inside `async` function, so lack of `return` is transformed into Promise<void>.resolve()

src/models/outbound/pispOTPValidate.model.ts
```diff
@@ -90,37 +91,35 @@ export class PISPOTPValidateModel
...
-        )
-        this.logger.push({ res }).info('ThirdpartyRequests.patchConsentRequests request sent to peer')
+        type PutResponseOrError = tpAPI.Schemas.ConsentsPostRequest & fspiopAPI.Schemas.ErrorInformationObject
+        const putResponse = message as unknown as PutResponseOrError
+        if (putResponse.errorInformation) {
+          this.data.errorInformation = putResponse.errorInformation
+        } else {
+          this.data.consent = { ...message as unknown as tpAPI.Schemas.ConsentsPostRequest }
+        }
+        return Promise.resolve()
```
eoln, line 115: we can drop this line too

src/handlers/outbound/consentRequests/{ID}/validate.ts
kleyow: What do you suggest the default timeout be?

kleyow: going to set it to 30 akin to the sdk-scheme-adapter for now.

src/models/outbound/pispOTPValidate.model.ts
kleyow: What do you suggest I pass in here?

src/handlers/outbound/consentRequests/{ID}/validate.ts
```diff
@@ -0,0 +1,76 @@
...
+  const consentRequestsRequestId = request.params.ID
+  const authToken = payload.authToken
+  const toDFSPParticipantId = payload.toParticipantId
+
+  // prepare config
+  const data: PISPOTPValidateData = {
+    currentState: 'start',
+    consentRequestsRequestId: consentRequestsRequestId,
+    authToken: authToken,
+    toDFSPParticipantId: toDFSPParticipantId
```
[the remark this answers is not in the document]
kleyow: Yep I have named it such because I thought `toAuthServiceParticipantId` would be needed, but I clarified that the PISP doesn't contact the auth-service anymore.

changed to `toParticipantId`

src/models/outbound/pispOTPValidate.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+  consent?: tpAPI.Schemas.ConsentsPostRequest
+  errorInformation?: tpAPI.Schemas.ErrorInformation
+}
+
+export interface OutboundOTPValidateData extends StateData {
+  authToken: string
+  toParticipantId: string
+}
+
+export interface OutboundOTPValidateConsentResponse {
```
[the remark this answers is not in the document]
kleyow: done here and throught for outbound interfaces.

src/handlers/outbound/consentRequests/{ID}/validate.ts
```diff
@@ -0,0 +1,76 @@
...
+    kvs: h.getKVS(),
+    pubSub: h.getPubSub(),
+    key: consentRequestsRequestId,
+    logger: h.getLogger(),
+    thirdpartyRequests: h.getThirdpartyRequests(),
+  }
+
+  const model: PISPOTPValidateModel = await create(data, modelConfig)
+
+  const result = await model.run()
```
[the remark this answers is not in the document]
kleyow: done

src/models/outbound/pispOTPValidate.model.ts
eoln: I would give it a try:
```suggestion
      await this.fsm.error(this.data.errorInformation)
```

kleyow: done.

src/interface/api-outbound.yaml
```diff
@@ -1829,6 +1824,21 @@ components:
         - toParticipantId
         - authToken
       additionalProperties: false
+    ConsentRequestsValidateState:
+      title: ConsentRequestsValidateState
+      description: State of POST consent requests validate
+      type: string
```
eoln, line 1830: I think there should be the enumeration
- errored
- success
- OTPisValid

src/models/outbound/pispOTPValidate.interface.ts
```diff
@@ -21,22 +21,46 @@
...
- * @param {string} args.consentRequestId - the consent request  id
- * @returns {string} - the pub/sub channel name
- */
-export function notificationChannel (consentRequestId: string) {
-  if (!consentRequestId) {
-    throw new Error('PISPConsentRequest.notificationChannel: \'consentRequestId\' parameter is required')
-  }
-  // channel name
-  return `consent_request_${consentRequestId}`
+export enum PISPOTPValidateModelState {
```
eoln, line 38: we can use here a type `OutboundAPI.Schemas.ConsentRequestsValidateState` so the enumeration values will come from API definition and there will be no need to redefine them here

test/step-definitions/consent-requests-validate-outbound.step.ts
```diff
@@ -0,0 +1,171 @@
...
+ ******/
+
+import { Server, ServerInjectResponse } from '@hapi/hapi'
+import { ServerAPI, ServerConfig } from '~/server'
+import { defineFeature, loadFeature } from 'jest-cucumber'
+import Config from '~/shared/config'
+import Handlers from '~/handlers'
+import index from '~/index'
+import path from 'path'
+import { NotificationCallback, Message, PubSub } from '../../src/shared/pub-sub';
```
eoln, line 34:
```suggestion
import { NotificationCallback, Message, PubSub } from '~/shared/pub-sub';
```

test/step-definitions/consent-requests-validate-outbound.step.ts
```diff
@@ -0,0 +1,171 @@
...
+import { RedisConnectionConfig } from '../../src/shared/redis-connection';
+import { thirdparty as tpAPI } from '@mojaloop/api-snippets';
+
+const apiPath = path.resolve(__dirname, '../../src/interface/api-outbound.yaml')
```
eoln, line 38:
```suggestion
const apiPath = path.resolve(__dirname, '~/interface/api-outbound.yaml')
```

test/integration/outbound/pisp-otp-validate.test.ts
```diff
@@ -0,0 +1,100 @@
...
+ ******/
+import axios from 'axios'
+import env from '../env'
+import { KVS } from '~/shared/kvs'
+import { RedisConnectionConfig } from '~/shared/redis-connection'
+import Config from '~/shared/config'
+import mockLogger from '../../unit/mockLogger'
+import { PISPOTPValidateModelState } from '~/models/outbound/pispOTPValidate.interface';
+
+describe.only('PISP OTP Validate', (): void => {
```
eoln, line 35:
```suggestion
describe('PISP OTP Validate', (): void => {
```

test/integration/outbound/pisp-otp-validate.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  const config: RedisConnectionConfig = {
+    host: Config.REDIS.HOST,
+    port: Config.REDIS.PORT,
+    logger: mockLogger(),
+    timeout: Config.REDIS.TIMEOUT
+  }
+  let kvs: KVS
+  const consentRequestsRequestId = 'ab049ac1-b936-4327-ab76-3e66e608beb8'
```
eoln, line 43: I would add here a comment that these values are used by TTK to simulate proper responses

eoln: Wouldn't it be better to use `authToken` value to switch between simulation cases in TTK instead of using `consentRequestId`? 

Then we can use `uuid()` function to generate id.

Unique id for every call is a part of the interface protocol,
and we didn't test this requirement when using hardcoded ones.

Using unique id will improve our test coverage

test/integration/outbound/pisp-otp-validate.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        ]
+      }
+      const consentRequestsResponse = await axios.patch(consentRequestsURI, consentRequestsRequest)
+      expect(consentRequestsResponse.status).toEqual(200)
+      expect(consentRequestsResponse.data.currentState).toEqual(PISPOTPValidateModelState.OTPIsValid)
+      expect(consentRequestsResponse.data.consent).toEqual(expectedResponse)
+    })
+  })
+
+  describe.only('/consentRequests/{ID}/validate: requestAction->errored error return', (): void => {
```
eoln, line 82:
```suggestion
  describe('/consentRequests/{ID}/validate: requestAction->errored error return', (): void => {
```

test/step-definitions/consent-requests-validate-outbound.step.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { Server, ServerInjectResponse } from '@hapi/hapi'
+import { ServerAPI, ServerConfig } from '~/server'
+import { defineFeature, loadFeature } from 'jest-cucumber'
+import Config from '~/shared/config'
+import Handlers from '~/handlers'
+import index from '~/index'
+import path from 'path'
+import { NotificationCallback, Message, PubSub } from '../../src/shared/pub-sub';
+import { RedisConnectionConfig } from '../../src/shared/redis-connection';
```
eoln, line 35:
```suggestion
import { RedisConnectionConfig } from '~/shared/redis-connection';
```

test/unit/models/outbound/pispOTPValidate.model.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const genericErrorResponse: fspiopAPI.Schemas.ErrorInformationObject = {
+      errorInformation: {
+        errorCode: '6000',
+        errorDescription: 'Generic thirdparty error'
+      }
+    }
+
+    beforeEach(async () => {
+      validateData = {
```
eoln, line 154: this part can be moved to `L132` as `const`

eoln: and we can drop `beforeEach` when empty

test/unit/models/outbound/pispOTPValidate.model.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        consentRequestsRequestId: '45789d6b-9423-4876-ab98-a12ff5b54195',
```
eoln, line 155: let define `'45789d6b-9423-4876-ab98-a12ff5b54195'` as independent `const` so it can be reused instead copy/pasting the value around

test/unit/handlers/outbound-otp-validate.test.ts
```diff
@@ -0,0 +1,92 @@
...
+      payload: validateRequest
+    }
+
+    const result = await ConsentRequestsIDValidate.patch(
+      {},
+      request as unknown as Request,
+      toolkit as unknown as StateResponseToolkit
+    )
+    expect(result.statusCode).toBe(500)
+  })
```
eoln, line 91: we should also check response value here

eoln (changes requested):
It's a big and a nice improved code! Thanks for addressing all my comments.
I've found few more glitches ;-)

src/models/outbound/pispOTPValidate.interface.ts
kleyow: How do I turn an interface into a type? or enum in this case?

test/step-definitions/consent-requests-validate-outbound.step.ts
kleyow: Is there a plugin to let vscode know path alias' so it imports using the alias instead of relative?

test/unit/handlers/outbound-otp-validate.test.ts
kleyow: the and model toolkit are mocked here, so I shouldn't need to check here?

test/step-definitions/consent-requests-validate-outbound.step.ts
kleyow: I don't think that works with path.resolve?

src/interface/api-outbound.yaml
kleyow: done.

test/integration/outbound/pisp-otp-validate.test.ts
kleyow: woops. fixed.

test/integration/outbound/pisp-otp-validate.test.ts
kleyow: done and switched to using authToken as the rule.

test/step-definitions/consent-requests-validate-outbound.step.ts
eoln: yeah, you are right -> `path.resolve` knows nothing about this

test/unit/handlers/outbound-otp-validate.test.ts
eoln: yeah, you are right!

test/step-definitions/consent-requests-validate-outbound.step.ts
eoln: I have some auto import plugin, but I always have to fix how it imports :-D

src/models/outbound/pispOTPValidate.interface.ts
eoln: - simply replace `interface` with `type` keyword
- to replace `enum` by type is little more complex -> you can use an alternate type operator ->  `type X = 'A' | '1'` -> but then you cannot use dot notation anymore -> you are using the values.

eoln: to use `ConsentRequestsValidateState` you need to reimport it from generated openapi.d.ts file into `interface/outbound/api-interface/index.ts` example https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/interface/outbound/api_interfaces/index.ts#L57 so I can use it here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/models/pispTransaction.interface.ts#L50

eoln: Maybe not complex but different ;-) For `enum addicted` persons using strict values instead enumType.dot notation would be strange, but TS doesn't allow you to assign invalid strict values what makes dot notation obsolete and too elaborative. And you can write much less code without dot notation.
Yes, you are right `PISPTransactionModelState` is the leftover from my last refactoring where I migrated from enums to autogenerated types & strict values - to be honest, I am considering making `StateData` from `persistent.model` a generic one, parametrized by state type so then we will have full type checking ;-)

kleyow: changed to raw strings using `currentState: 'start' | 'OTPIsValid' | 'errored',` instead. done.

src/models/outbound/pispOTPValidate.interface.ts
```diff
@@ -21,22 +21,40 @@
...
+  onValidateOTP: Method
+}
+
+export interface PISPOTPValidateModelConfig extends PersistentModelConfig {
+  pubSub: PubSub
+  thirdpartyRequests: ThirdpartyRequests
+  requestProcessingTimeoutSeconds: number
+}
+
+export interface PISPOTPValidateData extends StateData {
```
eoln, line 48: I think you can import `ConsentRequestsValidateState` from `src/interface/outbound/api_interfaces/index.ts` 

and use it as a generic parameter when extending StateData

like here: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/models/pispTransaction.interface.ts#L86
so you don't have to redefine currentState

```suggestion
export interface PISPOTPValidateData extends StateData<ConsentRequestsValidateState> {
```

src/models/outbound/pispOTPValidate.interface.ts
```diff
@@ -21,22 +21,40 @@
...
+  currentState: 'start' | 'OTPIsValid' | 'errored',
```
eoln, line 49: when using generic param for StateData - you are not forced to redefine currentState and use the source of truth delivered by autogenerated interfaces/types

src/models/outbound/pispOTPValidate.interface.ts
kleyow: Ah I didn't see that you already checked that in with a PR. Thought that it was something you were going to do. done.

eoln (approved):
great work done @kleyow - many thanks!

PR #80, opened 2021-04-09
feat: implement consentRequests( post and put) endpoints

sridharvoruganti:
related to #[2001](https://github.com/mojaloop/project/issues/2001)

### DIFF
--- a/src/handlers/inbound/consentRequests/{ID}.ts
+++ b/src/handlers/inbound/consentRequests/{ID}.ts
@@ -41,7 +41,6 @@ import inspect from '~/shared/inspect';
 import { PISPConsentRequestsModel } from '~/models/outbound/pispConsentRequests.model';
 import { Message } from '~/shared/pub-sub'
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
 async function patch (_context: unknown, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
   const payload = request.payload as tpAPI.Schemas.ConsentRequestsIDPatchRequest
   const consentRequestsRequestId = request.params.ID
@@ -84,9 +83,7 @@ async function patch (_context: unknown, request: Request, h: StateResponseToolk
 /**
  * Handles an inbound `PUT /consentRequests/{ID}` request
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
 async function put (_context: unknown, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
-  //const payload = request.payload as tpAPI.Schemas.ConsentRequestsIDPutResponseWeb
   const consentRequesttId = request.params.ID
 
   const channel = PISPConsentRequestsModel.notificationChannel(consentRequesttId)
--- a/src/handlers/inbound/consentRequests/{ID}.ts
+++ b/src/handlers/inbound/consentRequests/{ID}.ts
@@ -89,8 +89,11 @@ async function put (_context: unknown, request: Request, h: StateResponseToolkit
   const channel = PISPConsentRequestsModel.notificationChannel(consentRequesttId)
   const pubSub = h.getPubSub()
   // don't await on promise to resolve, let finish publish in background
-  pubSub.publish(channel, request.payload as unknown as Message)
-  h.getLogger().info(`Inbound received PUT /consentRequests/{ID} response and published to channel : ${channel}`)
+  setImmediate(async () => {
+    pubSub.publish(channel, request.payload as unknown as Message)
+    h.getLogger().info(`Inbound received PUT /consentRequests/{ID} response and published to channel : ${channel}`)
+  })
+
   return h.response().code(200)
 }
 
--- a/src/handlers/inbound/consentRequests/{ID}/error.ts
+++ b/src/handlers/inbound/consentRequests/{ID}/error.ts
@@ -29,26 +29,31 @@ import { Message } from '~/shared/pub-sub'
 import { StateResponseToolkit } from '~/server/plugins/state'
 import { PISPOTPValidateModel } from '~/models/outbound/pispOTPValidate.model';
 import { Enum } from '@mojaloop/central-services-shared';
+import { PISPConsentRequestsModel } from '~/models/outbound/pispConsentRequests.model'
 
 /**
  * Handles a inbound PUT /consentRequests/{ID}/error request
  */
 async function put (_context: any, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
-    // PUT /consentsRequests/{ID}/error is a response to PATCH /consentRequests
-    // when an OTP or secret failed to validate.
-    // We publish the request on the PISPConsentRequestModel
-    const payload = request.payload as fspiopAPI.Schemas.ErrorInformation
+  // PUT /consentsRequests/{ID}/error is a response to PATCH /consentRequests
+  // when an OTP or secret failed to validate.
+  // We publish the request on the PISPConsentRequestModel
+  const payload = request.payload as fspiopAPI.Schemas.ErrorInformation
 
-    const channel = PISPOTPValidateModel.notificationChannel(
-      request.params.ID
-    )
-    const pubSub = h.getPubSub()
+  const channel = PISPOTPValidateModel.notificationChannel(
+    request.params.ID
+  )
+  const pubSub = h.getPubSub()
 
-    // don't await on promise to resolve, let finish publish in background
-    pubSub.publish(channel, payload as unknown as Message)
-    return h.response({}).code(Enum.Http.ReturnCodes.OK.CODE)
+  // don't await on promise to resolve, let finish publish in background
+  pubSub.publish(channel, payload as unknown as Message)
+
+  const consentReqChannel = PISPConsentRequestsModel.notificationChannel(request.params.ID)
+  pubSub.publish(consentReqChannel, payload as unknown as Message)
+
+  return h.response({}).code(Enum.Http.ReturnCodes.OK.CODE)
 }
 
 export default {
-    put
+  put
 }
--- a/src/handlers/inbound/consentRequests/{ID}/error.ts
+++ b/src/handlers/inbound/consentRequests/{ID}/error.ts
@@ -39,18 +39,19 @@ async function put (_context: any, request: Request, h: StateResponseToolkit): P
   // when an OTP or secret failed to validate.
   // We publish the request on the PISPConsentRequestModel
   const payload = request.payload as fspiopAPI.Schemas.ErrorInformation
-
   const channel = PISPOTPValidateModel.notificationChannel(
     request.params.ID
   )
   const pubSub = h.getPubSub()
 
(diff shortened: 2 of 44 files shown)


kleyow:
For integration tests you can merge in master where I've skipped that test. The refactor of transactions has broken the test for now.

src/handlers/inbound/consentRequests/{ID}.ts
```diff
@@ -79,6 +80,21 @@ async function patch (_context: unknown, request: Request, h: StateResponseToolk
...
+/**
+ * Handles an inbound `PUT /consentRequests/{ID}` request
+ */
+async function put (_context: unknown, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
+  const consentRequesttId = request.params.ID
+
+  const channel = PISPConsentRequestsModel.notificationChannel(consentRequesttId)
+  const pubSub = h.getPubSub()
+  // don't await on promise to resolve, let finish publish in background
+  pubSub.publish(channel, request.payload as unknown as Message)
```
lewisdaly, line 92: Can we please use `setImmediate()` like the above to keep things consistent?

src/handlers/inbound/consentRequests/{ID}/error.ts
```diff
@@ -29,26 +29,31 @@ import { Message } from '~/shared/pub-sub'
...
+  const channel = PISPOTPValidateModel.notificationChannel(
+    request.params.ID
+  )
+  const pubSub = h.getPubSub()
 
-    // don't await on promise to resolve, let finish publish in background
-    pubSub.publish(channel, payload as unknown as Message)
-    return h.response({}).code(Enum.Http.ReturnCodes.OK.CODE)
+  // don't await on promise to resolve, let finish publish in background
+  pubSub.publish(channel, payload as unknown as Message)
```
lewisdaly, line 49: Can we please use `setImmediate()` like the above to keep things consistent?

src/interface/api-outbound.yaml
```diff
@@ -334,6 +334,58 @@ paths:
           $ref: '#/components/responses/501'
         '503':
           $ref: '#/components/responses/503'
+  /consentRequests:
```
lewisdaly, line 337: Why is this on the outbound interface? I think I'm missing something here... Are we reusing the same request but making it syncronous?

lewisdaly (changes requested):

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,281 @@
...
+          null as unknown as string,
+          null as unknown as string,
+          null as unknown as string,
+          Errors.MojaloopApiErrorCodes.TP_FSP_CONSENT_SCOPES_ERROR
+        )
+      }
+
+      if (!response.isValid) {
+        throw new Errors.MojaloopFSPIOPError(
+          null as unknown as string,
```
lewisdaly, line 101: Same as above.

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,281 @@
...
+          null as unknown as string,
+          null as unknown as string,
+          Errors.MojaloopApiErrorCodes.TP_NO_SUPPORTED_SCOPE_ACTIONS
```
lewisdaly, line 104: Can you really infer this error from `response.isValid`? That seems like a bit of a logical leap to me

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,281 @@
...
+      // throw error to stop state machine
+      throw error
+    }
+  }
+
+  async onStoreReqAndSendOTP (): Promise<void> {
+    const { request, response } = this.data
+
+    try {
+      if (response?.authChannels.includes('WEB')) {
```
lewisdaly, line 149: Can you please explain this to me? Is there a case where `authChannels` has 2 elements? Or do we validate for that elsewhere?

I would prefer something more readable and future proof like:

```ts
const channel = response.authChannels.pop();
switch (channel): {
  case 'WEB': ..
  case 'OTP':...
  default: ...
}
```

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,281 @@
...
+      }
+
+    } catch (error) {
+      this.logger.push({ error }).error('RequestIsValid -> success')
+      // throw error to stop state machine
+      throw error
+    }
+  }
+
+  reformatError (err: Error): Errors.MojaloopApiErrorObject {
```
lewisdaly, line 162: Does this belong elsewhere, like utils or something?

lewisdaly (changes requested):

src/handlers/inbound/consentRequests/{ID}.ts
sridharvoruganti: Ok fixed

src/handlers/inbound/consentRequests/{ID}/error.ts
sridharvoruganti: Ok fixed

src/interface/api-outbound.yaml
sridharvoruganti: Yes this outbound synchronous endpoint is for e2e testing purpose

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  async onValidateRequest (): Promise<void> {
+    const { request, toParticipantId } = this.data
+
+    try {
+      const response = await this.dfspBackendRequests.validateConsentRequests(request)
+
+      if (!response) {
+        throw new Errors.MojaloopFSPIOPError(
+          null as unknown as string,
```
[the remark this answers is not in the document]
sridharvoruganti: for some reason if back-end returns empty response, then will throw exception with specific error code 
we are throwing exception this way in case of validation errors 
Kevin used like that , so followed same pattern.
https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/models/inbound/dfspOTPValidate.model.ts#L108

src/models/inbound/dfspConsentRequests.model.ts
sridharvoruganti: when PISP sends  `POST /consentRequests` to DFSP, it validates that request scopes and responds with 
`PUT /consentRequests`or `PUT /consentRequests/{id}/error`
response contains  2 extra fields . 1 authChannels['WEB' or 'OTP']  2. authUri(only in case  of web)
these details returned from dfsp back-end or mojaloop-simulator 
here if else checks response authChannels[] and if it's web, returns authChannels, authUri otherwise only authChannels details to pisp
as there are only 2 scenarios I used if -else,  will  change it to switch  
`Is there a case where authChannels has 2 elements?  I am not sure about this, it's design decision`

src/models/inbound/dfspConsentRequests.model.ts
sridharvoruganti: Ok moved it to separate file shared/util

src/models/inbound/dfspConsentRequests.model.ts
sridharvoruganti: Ok now I have modified the code to forward back-end response as it is without those checks

src/models/inbound/dfspConsentRequests.model.ts
sridharvoruganti: As discussed backend response contains validation details

src/interface/api-outbound.yaml
lewisdaly: I don't love that we are reusing the resource names, but that's a fight for another day...

src/models/inbound/dfspConsentRequests.model.ts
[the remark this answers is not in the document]
lewisdaly: As discussed please write a wrapper for this maybe in utils, or fix upstream.

src/models/inbound/dfspConsentRequests.model.ts
lewisdaly: Yep, all good.

src/models/inbound/dfspConsentRequests.model.ts
lewisdaly: I'm still seeing the `if else` block..

sridharvoruganti: can you pls refresh,  I see new changes
https://github.com/mojaloop/thirdparty-scheme-adapter/blob/4cac6a7f00b9e7a5e9e102fbfb0013dcd1038a6c/src/models/inbound/dfspConsentRequests.model.ts

src/models/inbound/dfspConsentRequests.model.ts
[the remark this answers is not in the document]
sridharvoruganti: for now I modified that code to `throw new Errors.MojaloopFSPIOPError('', '', '', Errors.MojaloopApiErrorCodes.TP_FSP_CONSENT_SCOPES_ERROR) `
I think it looks fine now,  implementing wrapper or fix upstream require changes in sdk-standard-components.
will do those changes in a separate PR.

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,243 @@
...
+        case 'WEB': {
+          await this.dfspBackendRequests.storeConsentRequests(request)
+          break
+        }
+        case 'OTP': {
+          await this.dfspBackendRequests.sendOTP(request)
+          break
+        }
+        default: {
+          console.log(`Invalid authChannel ${channel}`)
```
lewisdaly, line 144: This should throw an error.

lewisdaly (changes requested):
one last little change...

src/models/inbound/dfspConsentRequests.model.ts
sridharvoruganti: ok

src/models/inbound/dfspConsentRequests.model.ts
```diff
@@ -0,0 +1,243 @@
...
+        case 'WEB': {
+          await this.dfspBackendRequests.storeConsentRequests(request)
+          break
+        }
+        case 'OTP': {
+          await this.dfspBackendRequests.sendOTP(request)
+          break
+        }
+        default: {
+          console.error(`Invalid authChannel ${channel}`)
```
lewisdaly, line 144: I really think this should throw an error here.

lewisdaly: We probably also shouldn't log to console... we have a logger, right?

sridharvoruganti: ok done, instead of just logging, throwing new error in case of invalid auth channel

lewisdaly: cool

lewisdaly (approved):

PR #81, opened 2021-04-12
doc: create initial linking seq diagram

kleyow:
I'm thinking the account linking flow can be broken up into distinct models instead of one giant model.
Where the flow breaks into a new model every time the PISP needs user input.

Resulting into these 
PISPPrelinkingModel (handles 1.1 Prelinking)
User needs to select DFSP.

PISPDiscoveryModel (handles 1.2 Discovery. Currently named `OutboundAccountsModel`)
DFSPDiscoveryModel (handles 1.2 Discovery. Currently named `InboundAccountsModel`)
User needs to select account.

PISPRequestConsentModel (handles 1.3 Discovery)
DFSPRequestConsentModel (handles 1.3 Discovery)
User needs to complete web flow.
User needs to wait and enter OTP.
- :thinking: but where do we store the `consentRequestId` until the user has completed those steps...
   Do we make the MobileApp/Backend store it until an Authentication attempt...? and how...?
- We might need to make this apart of the `AuthenticationConsentCredentialModel` so that `consentRequestId` is in memory but user introduces so much unknown in and this would be one GIANT state machine.

No more user input is needed at this point. Complete account linking in one go.
PISPAuthenticationConsentCredentialModel (handles 1.4 Authentication throught to 1.6. Credential registration)
(Some logic handled by DFSPOTPValidateModel. Refactor to handle 1.4-1.6)
DFSPAuthenticationConsentCredentialModel (handles 1.4 Authentication throught to 1.6. Credential registration)
(Some logic handled by DFSPOTPValidateModel. Refactor to handle 1.4-1.6)

That being said. Here is the sequence diagram for Pre-linking assuming my above approach is agreed upon.


eoln:
@kleyow I suggest taking a look: https://raw.githubusercontent.com/mojaloop/pisp/master/docs/out/linking/2-request-consent-web.svg

especially SIM-1-3 steps are dedicated to storing consentRequest details at simulator for later use in SIM-13

eoln:
I would prefer one workflow with two side models (PISP/DFSP) for linking. 

Where PISPLinkingModel side is realized in phases in a similar way as we do for transfer (PISPTransactionModel: lookup, initiate, approve). Every phase is initiated on User's action on Outgoing. 

And DFSPLinkingModel where DFSP part of linking workflow is implemented which reacts on events coming from switch on Ingoing.

kleyow:
> I would prefer one workflow with two side models (PISP/DFSP) for linking.
> 
> Where PISPLinkingModel side is realized in phases in a similar way as we do for transfer (PISPTransactionModel: lookup, initiate, approve). Every phase is initiated on User's action on Outgoing.
> 
> And DFSPLinkingModel where DFSP part of linking workflow is implemented which reacts on events coming from switch on Ingoing.

Is there any advantage to having it be 2 big linking models vs breaking at points where a user needs to input? 

My thought process was instead of having a model sitting in memory waiting for a user to finish some action like selecting a DFSP to fetch account for it's better to break that into it's own model and have it resolve.

- User may take a long time to complete their action
- User might not complete the action and abort.

@eoln

lewisdaly:
As I said in the call (and after doing some additional thinking) I also prefer one model.

> My thought process was instead of having a model sitting in memory waiting for a user to finish some action...
This isn't really an issue. We use redis, redis takes care of this for us. We can always time out a state if need be after a day or two.

Another benefit for keeping one big model is that we have very well defined relationships between the parts of the model. Relying on multiple, small models defeats many of the purposes of using a state machine. Smaller models also makes it more difficult to add logic at a later date, since new logic may interact in unexpected ways to the implicit relationships between those smaller models.

kleyow:
:thumbs up:

PR #83, opened 2021-04-13
doc: create initial linking seq diagram 

kleyow:


kleyow:
Think this is a good stopping point before continuing. 
Things to note. 

- I think we'll need to create a uuid at the pre-linking step since this would all be one state machine.
  That uuid will serve as the key value and `consentRequestId` in the future.
- The DFSPLinkingModel will also need an ID that it works off of to work in tandem PISPLinkingModel.
  I think this would need to be passed in the discovery phase? Not sure how that would look since it's a GET request.
  
Though I'm not sure that will work in 1.6 Credential Registration, as the requests don't contain the `consentRequestId` that we can load the Model from KVS. 

Honestly not sure how we linking these phases to each other when some miss a common link :thinking:. That's something I might figure out in write the later phases.

eoln:
@kleyow PISP & DFSP sides are having independent states from each other. It is worth pointing out that PISPLinkingModel will be run at PISP infrastructure where DFSPLinkingModel will reside be at DFSP side. Both of them are using Redis to store state, but we have to treat them as totally isolated storages, so we can't pass any state mutation between DFSP/PISP via redis, only via API: `scheme adapters have always independent instances`

Handling `1.6. Credential Registration` is the responsibility of PISPLinkingModel, so when POST /consent at inbound is received from DFSP in request payload you are having `consentRequestId` so in inbound handler you can load the model using it and execute the `run` command. I think it is antipattern to think about these state machine models as dedicated only to Inbound/Outbound service -> it looks better for me when I think about them as a middleman who reacts on events sourced from Outgoing/Ingoing handlers

eoln:
@kleyow I've just discovered you are talking about uuid for prelinking to get the list of DFSP which would like to cooperate with PISP -> `GET /linking/dfsps` this query can be stateless -> PISP will store received list of valid DFSP internally to present it later to the user. For stateless query, there is no need to save the PISPLinkingModel state.

eoln:
@kleyow I think we can drop `pre-linking cached` it brings complications, more code, and the risk we are having cache improperly invalidated.

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+activate Switch
+Switch --> PISP_LM: 202 Accepted
+deactivate PISP_LM
+
+Switch -> DFSP_TP_IN: POST /consentRequests
+activate DFSP_TP_IN
+DFSP_TP_IN --> Switch: 202 Accepted
+deactivate Switch
+
+DFSP_TP_IN -> DFSP_TP_IN: const model = await loadFromKVS()
```
eoln, line 353: we are creating DFSPLinkinigModel here, we will call `saveToKVS` after receiving response form `fsm.requestConsent` transition (in `run` implementation)

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+    FSP does not trust PISP callback URI
+end note
+
+activate DFSP_TP_IN
+DFSP_TP_IN -> DFSP_TP_IN: const model = await loadFromKVS()
+activate DFSP_LM
+DFSP_TP_IN -> DFSP_LM: model.requestConsent()
+DFSP_LM -> DFSP_LM: DFSPBackendRequests.validateConsentRequest()
+
+DFSP_TP_IN --> Switch: 202 Accepted
```
eoln, line 414: let move it L349 before `DFSPBackendRequests.validateConsentRequest()`

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+note over PISP, DFSP_LM
+  ERROR_SCENARIO:
+    FSP does not find scopes suitable |
+    FSP does not support any requested authentication channels |
+    FSP does not support any requested scope actions |
+    FSP does not trust PISP callback URI
+end note
+
+activate DFSP_TP_IN
+DFSP_TP_IN -> DFSP_TP_IN: const model = await loadFromKVS()
```
eoln, line 409: create not load - same like for happy scenario

kleyow:
> think we can drop pre-linking cached phase. It brings complications, more code, and the risk we are having cache improperly invalidated.

Sure thing.

> I've just discovered you are talking about uuid for prelinking to get the list of DFSP which would like to cooperate with PISP -> GET /linking/dfsps this query can be stateless -> PISP will store received list of valid DFSPs internally to present it later to the user. For stateless query, there is no need to save the PISPLinkingModel state.

This would apply for `GET linking/accounts/dfspa/username1234` as well, yes? If they these two phases are stateless then what was the push for the whole linking process to be a singular state machine. Since these operations can exist in isolation?



> @kleyow PISP & DFSP sides are having independent states from each other. It is worth pointing out that PISPLinkingModel will be run at PISP infrastructure where DFSPLinkingModel will reside at DFSP side. Both of them are using Redis to store state, but we have to treat them as totally isolated storages, so we can't pass any state mutation between DFSP/PISP via redis, only via API: `scheme adapters have always independent instances` Moreover, in the future, it could be possible that DFSPLinkingModel for our scheme adapter is talking to DFSP custom implementation of PISP linking protocol - so any assumptions about the other side are limited only to API interfaces.
> 
> Handling `1.6. Credential Registration` is the responsibility of PISPLinkingModel, so when POST /consent at inbound is received from DFSP in request payload you are having `consentRequestId`. So in Inbound handler you can load the model using it and execute the `run` command.
> 
> I think it is antipattern to think about these state machine models as dedicated only to Inbound/Outbound service -> it looks better for me when I think about them as a middleman who reacts on events sourced from Outgoing/Ingoing handlers
> 
> Using this approach `consentRequestId` is the only common link between these two LinkingModels.

I'll talk about this in our daily. This is exactly my thinking so I'm not sure where the diagram is implying that each side connected other than through the API.

> Handling 1.6. Credential Registration is the responsibility of PISPLinkingModel, so when POST /consent at inbound is received from DFSP in request payload you are having consentRequestId. So in Inbound handler you can load the model using it and execute the run command.

The problem begins at the `PUT /consents/123` from the PISP -> DFSP in 1.6.2. The body has no `consentRequestId` anymore so when the DFSP recieves that on the inbound it would have no key to pull the model from storage.

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+    { accountNickname: "SpeXXXXXXXXnt", id: "dfspa.username.5678", currency: "USD" }
+  ],
+  callbackURI: 'pisp-app://callback'
+}
+end note
+
+activate PISP
+activate PISP_TP_OUT
+
+PISP_TP_OUT -> PISP_TP_OUT: const model = await loadFromKVS()
```
eoln, line 480: create not load - same like for Web

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+activate Switch
+Switch --> PISP_LM: 202 Accepted
+deactivate PISP_LM
+
+Switch -> DFSP_TP_IN: POST /consentRequests
+activate DFSP_TP_IN
+DFSP_TP_IN --> Switch: 202 Accepted
+deactivate Switch
+
+DFSP_TP_IN -> DFSP_TP_IN: const model = await loadFromKVS()
```
eoln, line 517: create not load - same like for Web

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+DFSP_TP_IN -> DFSP_LM: model.requestConsent()
+DFSP_LM -> DFSP_LM: DFSPBackendRequests.validateConsentRequest()
+
+activate DFSPAuthorizeSimulator
+DFSP_LM -> DFSPAuthorizeSimulator: ""POST /store/consentRequests/6789""
+DFSPAuthorizeSimulator -> DFSPAuthorizeSimulator: store consentRequest details
+DFSPAuthorizeSimulator -> DFSP_LM: 201 Created
+deactivate DFSPAuthorizeSimulator
+
+DFSP_TP_IN --> Switch: 202 Accepted
```
eoln, line 528: move it before L548: `DFSPBackendRequests.validateConsentRequest()`

docs/sequence/PISPLinkingApi.puml
```diff
@@ -0,0 +1,654 @@
...
+    FSP does not trust PISP callback URI
+end note
+
+activate DFSP_TP_IN
+DFSP_TP_IN -> DFSP_TP_IN: const model = await loadFromKVS()
+activate DFSP_LM
+DFSP_TP_IN -> DFSP_LM: model.requestConsent()
+DFSP_LM -> DFSP_LM: DFSPBackendRequests.validateConsentRequest()
+
+DFSP_TP_IN --> Switch: 202 Accepted
```
eoln, line 580: move before L606: DFSPBackendRequests.validateConsentRequest()

eoln (changes requested):
I see it going in a good direction. 
- @lewisdaly I am against a GIGANT model for both sides. Let have two one for PISP side and another for DFSP side of the flow - like in Transaction flow
- if we are sending 202 Accept back to switch we should do it ASAP, and invoke background task after.

lewisdaly:
> @lewisdaly I am against a GIGANT model for both sides. Let have two one for PISP side and another for DFSP side of the flow - like in Transaction flow

Yeah  I agree

kleyow:
@lewisdaly I am against a GIGANT model for both sides. Let have two one for PISP side and another for DFSP side of the flow - like in Transaction flow

I'm confused...is that not what the sequence diagram suggests?
It's two models. One for each side. Which happen to both be gigant...

PR #91, opened 2021-05-06
doc: add endpoint to handle passing of credential and fix typos

kleyow:


kleyow:
I'm going to merge this in since I had a verbal +1 after standup.

PR #95, opened 2021-05-14
feat: implement pisp prelinking flow

kleyow:


src/models/outbound/pispPrelinking.model.ts
```diff
@@ -0,0 +1,200 @@
...
+ Gates Foundation organization for an example). Those individuals should have
+ their names indented and be marked with a '-'. Email address can be added
+ optionally within square brackets <email>.
+ * Gates Foundation
+ - Name Surname <name.surname@gatesfoundation.com>
+
+ - Kevin Leyow - kevin.leyow@modusbox.com
+ --------------
+ ******/
+
```
eoln, line 27: wouldn't it be simpler to use here `Async2SyncModel`? 
example how to generate the model: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/test/unit/models/a2s.model.test.ts

src/handlers/inbound/services/{ServiceType}.ts
```diff
@@ -0,0 +1,58 @@
...
+async function put (_context: unknown, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
+  const serviceType = request.params.ServiceType
+
+  if (serviceType == ServiceType.THIRD_PARTY_DFSP) {
+    const channel = PISPPrelinkingModel.notificationChannel(serviceType)
+    const pubSub = h.getPubSub()
+
+    // don't await on promise to resolve, let finish publish in background
+    setImmediate(async () => {
+      pubSub.publish(channel, request.payload as unknown as Message)
```
eoln, line 48: instead of using pubSub.publish - I would use `deferredJob.trigger` method instead -> then all handlers will only depend on your model -> the purest way would be to add a static method to your model class which wraps around deferredJob.trigger invocation -> then your pub/sub code will be used only in your model

test/step-definitions/linking-providers-outbound.step.ts
```diff
@@ -0,0 +1,157 @@
...
+
+defineFeature(feature, (test): void => {
+  let server: Server
+  let response: ServerInjectResponse
+
+  afterEach((done): void => {
+    jest.resetAllMocks()
+    jest.resetModules()
+    server.events.on('stop', done)
+    server.stop()
```
eoln, line 118: let stop server ASAP so the unit tests will not hung-up
```suggestion
    server.stop({ timeout:0 })
```

test/step-definitions/services-inbound.step.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  // tests seem to not like the server booting up/down between tests.
+  // so we prepare a server for all tests in the feature
+  beforeAll(async (): Promise<void> => {
+    server = await prepareInboundAPIServer()
+  })
+
+  afterAll(async (done): Promise<void> => {
+    server.events.on('stop', done)
+    server.stop()
```
eoln, line 64:
```suggestion
    server.stop({ timeout:0 })
```

eoln (changes requested):

test/step-definitions/services-inbound.step.ts
kleyow: done.

test/step-definitions/linking-providers-outbound.step.ts
kleyow: done

src/handlers/inbound/services/{ServiceType}.ts
kleyow: done

src/models/outbound/pispPrelinking.model.ts
kleyow: talked about this in standup. going to leave it using the persistant model for now.

src/handlers/inbound/services/{ServiceType}/error.ts
```diff
@@ -0,0 +1,57 @@
...
+async function put (_context: any, request: Request, h: StateResponseToolkit): Promise<ResponseObject> {
+  const payload = request.payload as fspiopAPI.Schemas.ErrorInformation
+  const channel = PISPPrelinkingModel.notificationChannel(
+    request.params.ServiceType
+  )
+  const pubSub = h.getPubSub()
+
+  // don't await on promise to resolve, let finish publish in background
+  setImmediate(async () => {
+    pubSub.publish(channel, payload as unknown as Message)
```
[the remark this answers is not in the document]
kleyow: done

src/handlers/inbound/services/{ServiceType}.ts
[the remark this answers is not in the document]
kleyow: done

eoln (approved):

PR #102, opened 2021-06-04
test: add robust DFSP side integration tests

kleyow:


eoln (changes requested):
there shouldn’t be a need for separate client - because there are already sepearated in state-plugin: https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/server/plugins/state.ts#L74
the kvs&pubSub instances are then connected and exposed by state decorator here:
|https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/server/plugins/state.ts#L160
https://github.com/mojaloop/thirdparty-scheme-adapter/blob/master/src/server/plugins/state.ts#L164-L165

kleyow:
The KVS and PubSub being separated is fine, the PubSub is the one that is needing to be seperated. In retro my comments might be a bit misleading and my solution not the best.
The issue is that the `RedisConnection` only creates one client, the `PubSub` child class is using that one client for both publish and subscription. That client once a `subscribe` operation is run onto it, it becomes a Subscription based client unable to perform other non-subscription requests.
The right solution I should have made was create a `PubSubRedisConnection` so that I don't change `KVS`'s parent class. Or create a two instances of PubSub in our state decorator...:expressionless: which would have been the easier solution, realizing now. But someone could accidently write code that uses the wrong instance.
Right now all our handlers are using the same PubSub instance i.e h.getPubSub() to create our models and the same instance is passed into triggerWorkflow()  so our models are subscribing on that client and triggerWorkflow() is publishing on the same client.
I've push a much simpler example. What the integration test is doing is the right thing which is not what we are doing in our handlers.
https://github.com/mojaloop/thirdparty-scheme-adapter/pull/103/files#diff-c3e410631c37fd1f99d1046d6c9de1e1cc34e7ac4d49cfd2cd765937df873525R43

kleyow:
@lewisdaly can't nav to link. private repo?

> In this instance, I separate out a subscriber and publisher, which own their own redis connections.

tl;dr we weren't seperating out the redis connection into multiple clients. we had one redis connection with one client. pub and sub calls were being made on that one client. redis doesn't like that. 

so I updated the redis connection to have two clients. one for sub. one for pub/get/set/etc operations.

lewisdaly:
oops, here's the relevant snippet:

```ts
 // a redis connection can't be both a publisher and subscriber at the same time
const subscriber = new PubSub(redisConnectionConfig)
await subscriber.connect()
const publisher = new PubSub(redisConnectionConfig)
await publisher.connect()
const globalState = new GlobalClientState(kvs, subscriber, publisher)
const connectionWatcher = new ConnectionWatcher(globalState, 30000);
connectionWatcher.startWatching()
```

kleyow:
Yes that solution is what I refer to here. 
```
Or create a two instances of PubSub in our state decorator...:expressionless: which would have been the easier solution, realizing now. But someone could accidently write code that uses the wrong instance.
```
I chose to amend the parent `RedisConnection` class to support the `PubSub` use case instead of creating two `PubSub` instances.

lewisdaly:
>someone could accidently write code that uses the wrong instance.

Is that a real concern? If one is called `publisher` and the other `subscriber`, then how could they get it wrong? I feel like my proposed solution is much cleaner and generalizes better than your proposal.

kleyow:
```
export interface StateResponseToolkit extends ResponseToolkit {
  getKVS: () => KVS
  getPubSub: () => PubSub
  getLogger: () => SDKLogger.Logger
  getMojaloopRequests: () => MojaloopRequests
  getThirdpartyRequests: () => ThirdpartyRequests
  getWSO2Auth: () => WSO2Auth
  getPISPBackendRequests: () => PISPBackendRequests
  getDFSPBackendRequests: () => DFSPBackendRequests
  getSDKOutgoingRequests: () => SDKOutgoingRequests
  getDFSPId: () => string
  getAuthServiceParticipantId: () => string
}
```

I feel like since they are obsured in the toolkit and being passed around in models and fuctions it might be confusing to keep track of what instance you are working with. 

I'll switch to the two instance solution to get this over the line and also since I have worries that keeping two clients open on a connection/class that might not need it (in the case of the KVS) might be wasteful.

lewisdaly:
Thanks Kevin, sorry to be a pain about this.

kleyow:
@lewisdaly changes made

test/integration/inbound/dfsp-linking.test.ts
```diff
@@ -0,0 +1,1319 @@
...
+    const axiosConfig = {
+      headers: {
+        'Content-Type': 'application/json',
+        'FSPIOP-Source': 'switch',
+        Date: 'Thu, 24 Jan 2019 10:23:12 GMT',
+        'FSPIOP-Destination': 'dfspA'
+      }
+    }
+    describe('Inbound POST /consentRequests from PISP should create DFSP Linking and start the workflow', (): void => {
+      it('should send back PUT /consentRequests/{ID} containing the specified auth channel', async (): Promise<void> => {
```
lewisdaly, line 59: So this test passes without `--runInBand=true`?

test/integration/inbound/dfsp-linking.test.ts
```diff
@@ -0,0 +1,1319 @@
...
+
+        const response = await axios.post(scenarioUri, payload, axiosConfig)
+
+        // Switch should return Accepted code to DFSP
+        expect(response.status).toEqual(202)
+
+        // wait a bit for the DFSP adapter to process the request
+        await new Promise(resolve => setTimeout(resolve, 200));
+
+        // check that the DFSP has sent a PUT /consentRequests/{ID} to the PISP
```
lewisdaly, line 110: Nice, this looks really clean and elegant

test/integration/inbound/dfsp-linking.test.ts
lewisdaly: Ah I see, they are in separate `describe` blocks. Looks good.

test/integration/inbound/dfsp-linking.test.ts
```diff
@@ -0,0 +1,1319 @@
...
+        )
+      })
+    })
+  })
+
+  describe('DFSP dfspBackendRequests.validateConsentRequests returns a HTTP error', (): void => {
+    describe('POST /consentRequests should create DFSP Linking and start the workflow', (): void => {
+      it('should send a PUT /consentRequests/{ID}/error to PISP', async (): Promise<void> => {
+        /*
+        the `consentRequestId` is the variable that changes the course of
```
lewisdaly, line 585: nice!

test/integration/inbound/dfsp-linking.test.ts
```diff
@@ -0,0 +1,1319 @@
...
+              errorCode: "7213",
+              errorDescription: "Consent is invalid"
+            }
+          }
+        )
+      })
+    })
+  })
+
+  describe('DFSP recieves error from ALS after sending the auth-service PISP PENDING credential', (): void => {
```
lewisdaly, line 1122: I'm not sure I followed this error case. What is supposed to happen after `Inbound PUT /consents/{ID} signed credential from PISP`? Are we meant to be testing when the auth-service fails to register the credential with the ALS?

test/integration/inbound/participants.test.ts
```diff
@@ -31,7 +31,7 @@ import axios from 'axios'
 import env from '../env'
 import mockLogger from '../../unit/mockLogger'
 
-describe('DFSP Inbound', (): void => {
+describe.skip('DFSP Inbound', (): void => {
```
lewisdaly, line 34: Is this supposed to be skipped? Or does it need fixing?

lewisdaly (changes requested):
Wow, mammoth work here @kleyow. Well done!

I just have a couple questions about the final integration test case here. 

There was also a test suite that looks like the `.skip` needs to be removed (or explained).

test/integration/inbound/participants.test.ts
kleyow: skipped for now. was part of the mistake ALS handling. I chose to skip instead of completely remove the file since I might need parts of these tests for the THIRD_PARTY_LINK PHASE in the future.

kleyow:
Not at my desk atm so I'll push changes in a bit.

test/integration/inbound/dfsp-linking.test.ts
kleyow: deleted.

kleyow:
@lewisdaly believe I've addressed all comments.

lewisdaly (approved):

PR #108, opened 2021-07-27
feat: add consentId override and allow pisp to specify scope actions

kleyow:
…ions


lewisdaly (commented):
I'm not sure I have enough context here to review this. Can you explain why we might want this feature? 

I'm guessing its for tests, but I wonder if this functionality would be better suited to the simulators. I worry about giving DFSPs environment variables that allow them to compromise their systems... making ConsentID deterministic would compromise the security of the credential.

kleyow:
> I'm not sure I have enough context here to review this. Can you explain why we might want this feature?

It's only for tests so I can control `consentId` to be the validated credential you gave me since that needs to be correct for `deriveChallenge()` in e2e tests.

I struggled with the next best option of making an additional `dfsp-backend` request for the sole purpose of asking the dfsp to give the adapter a `consentId` uuid to use (which would solve us being able to control `consentId` in tests as a side-benefit). 

Thats just another endpoint the dfsp needs to integrate with which most likely on their backend will be returning a  `uuidv4()` string. I couldn't think of benefits a new `dfsp-backend` endpoint would provide. Maybe you have some fresh perspective.

@lewisdaly

lewisdaly:
Hmm. I see now. Thanks @kleyow it does seem like the best, easiest way for now. Can you please add a warning to the description in `config.ts`?  And also a `Logger.warn` message somewhere that can alert someone who has set this variable that they shouldn't be using it in production?

kleyow:
@lewisdaly done.

lewisdaly (approved):

PR #110, opened 2021-08-04
feat(outbound): update the outbound interface

lewisdaly:
- remove unused resources
- simplify the error responses
- add request and response examples


src/interface/outbound/api_template/api-outbound-template.yaml
```diff
@@ -1,38 +1,36 @@
+# Note: I've manually edited the `responses` section to replace all errors with a single
```
kleyow, line 1: Oh nice, did not know that. I'll make sure to update the snippets with that another time.

kleyow (approved):

PR #112, opened 2021-08-20
feat: add req in account linking so DFSP can store consent details

kleyow:


src/models/inbound/dfspLinking.interface.ts
```diff
@@ -81,6 +81,12 @@ export interface BackendGetScopesResponse {
   scopes: tpAPI.Schemas.Scope[]
 }
 
+export interface BackendStoreValidatedConsentRequest {
```
lewisdaly, line 84: Is there a reason you're not also sending the credential along here?

lewisdaly (commented):
Looks good to me as far as I understand this service... just 1 comment.

src/models/inbound/dfspLinking.interface.ts
kleyow: I can. Do you want me to just send the whole credential?

lewisdaly: Yeah I think so... that way the DFSP has it for auditing purposes or if they want to check the credential themselves.

lewisdaly (approved):

PR #114, opened 2021-08-25
feat(dfsp-transaction): get consent context

lewisdaly:


test/unit/models/dfspTransaction.model.test.ts
```diff
@@ -447,8 +453,15 @@ describe('DFSPTransactionModel', () => {
 
     it('should throw if transactionRequestRequest is not valid', async (done) => {
       mocked(modelConfig.kvs.set).mockImplementationOnce(() => Promise.resolve(true))
-      mocked(modelConfig.dfspBackendRequests.validateThirdpartyTransactionRequest)
-        .mockImplementationOnce(() => Promise.resolve({ isValid: false }))
+      mocked(modelConfig.dfspBackendRequests.validateThirdpartyTransactionRequestAndGetContext)
+        .mockImplementationOnce(() => Promise.resolve({
```
kleyow, line 457: I believe this will need to retrieve the challenge used to register the credential with the `auth-service`
https://github.com/mojaloop/auth-service/blob/master/src/domain/consents/ID.ts#L55 

The challenge is used as an extra check atm in the `auth-service`. I believe we talked about this. 
Wheter we have that extra check or remove it is fine with me. @lewisdaly

src/shared/dfsp-backend-requests.ts
```diff
@@ -167,10 +206,10 @@ export class DFSPBackendRequests extends HttpRequests {
     return this.post(this.validateAuthTokenPath, validateRequest)
   }
 
-  // validate ThirdpartyTransactionRequest
-  async validateThirdpartyTransactionRequest (
+  // validate ThirdpartyTransactionRequest and get context we will need to complete the transaction
```
kleyow, line 209: I'm fine with coupling the validate/lookup logic but you'll need to update the schema in `mojaloop-simulator` and update integration tests.

https://github.com/mojaloop/mojaloop-simulator/pull/97/files?file-filters%5B%5D=.yaml#diff-eb2a495bae6f23905b017eb7fea80e6920cb1c2356c450bc933ac38914bfb564R807

I personally don't mind if you just use `type: object` in `mojaloop-simulator` since I feel like we need to improve on our backend api and the request/response structures.

kleyow (changes requested):
2 comments. 
integration tests and backend request schema.
retrieving the registration credential challenge as part of the response.

src/shared/dfsp-backend-requests.ts
lewisdaly: Yep, I was thinking about this as well... Integration tests in this repo, or in the mojaloop-simulator repo?

test/unit/models/dfspTransaction.model.test.ts
lewisdaly: Hmm I have no idea why that code exists. Maybe I misunderstood you the other day, but there's no reason we should need the challenge from the credential registration step to verify a transaction.

src/shared/dfsp-backend-requests.ts
kleyow: This repo.

test/unit/models/dfspTransaction.model.test.ts
kleyow: Ok. No problem. That if statement will need to be removed then.

lewisdaly: Yeah for sure.

src/shared/dfsp-backend-requests.ts
lewisdaly: Ah I see. I forgot to look at the `pisp/master` branch on the mojaloop-simulator so I didn't follow your most recent changes

PR #120, opened 2021-09-06
feat(puml): update the PISP Transaction sequence to use `/tpr/verifications` calls

lewisdaly:
- chore(ci): reenables license check, fix bad env variables causing failures


kleyow (approved):

lewisdaly:
Thanks @kleyow !

PR #123, opened 2021-09-17
feat(linkingModel): deterministic test consentids

lewisdaly:
For more than 1 end to end test to run, we need a way to know ahead of time what the `consentId` will be. We previously did that with an environment variable, but I expanded this to allow us to specify more than 1 consentId, based on the consentRequestId


kleyow (approved):

kleyow:
Grrr...I think in the future I'll just an opt-in backend endpoint for a dfsp to specify endpoints, so we can have the best of both worlds

lewisdaly:
Yeah I think that's a more robust approach.

PR #142, opened 2022-05-17
feat!: update nodejs to lts

kleyow:
These peer-dependencies add some bit of annoyance managing packages...probably due to my lack of understanding.
A lot of the `central-services-*` libraries want different versions of peer-dependencies and npm v7 doesn't like that. So I have use `npm install --legacy-peer-dep`. 

Example `mojaloop/central-services-health` wants `mojaloop/central-services-logger@10.6.1` while
`mojaloop/central-services-shared` wants `mojaloop/central-services-logger@10.6.2` as laid out by their `peerDependencies`.

Is the solution to that to make sure all `central-services-*` are insync...?
What about libraries we don't maintain like `ajv-keywords` with its own peer `ajv` which differs from `central-services-*`?

Any npm grand master wizards chime in?


elnyry-sam-k:
Thanks @kleyow .. Looks like we need to address cs-shared and cs-health libraries to be in sync.. will discuss

elnyry-sam-k:
Also, node LTS v16.14.2 (central services logger PR here: https://github.com/mojaloop/central-services-logger/pull/37/files)

kleyow:
> Also, node LTS v16.14.2 (central services logger PR here: https://github.com/mojaloop/central-services-logger/pull/37/files)

Ah. The latest LTS is 16.15 but I guess we should just pick a LTS version and remain consistent throughout.

elnyry-sam-k (approved):

elnyry-sam-k:
> > Also, node LTS v16.14.2 (central services logger PR here: https://github.com/mojaloop/central-services-logger/pull/37/files)
> 
> Ah. The latest LTS is 16.15 but I guess we should just pick a LTS version and remain consistent throughout.

After discussion with Miguel, lets go with LTS 16.15.x :-). thanks Kevin!

PR #147, opened 2022-06-03
feat!: add ws client to pull updated configuration

kleyow:
- Ported over websocket client code from sdk-scheme-adapter
- Added feature to query websocket server for configuration on startup
- Added feature that will restart inbound/outbound servers on websocket configuration change message
- Conform config to camelcase/sdk-scheme-adapter since the client has merging config logic that depends on keys being the same
- Add parent server class that runs all servers/clients akin to sdk-scheme-adapter


config/development.json
```diff
@@ -1,9 +1,13 @@
...
-  "INBOUND": {
-    "PORT": 4005,
-    "HOST": "0.0.0.0",
-    "PISP_TRANSACTION_MODE": false,
-    "TLS": {
+  "control": {
+    "mgmtAPIWsUrl": "127.0.0.1",
+    "mgmtAPIWsPort": 4010
+  },
+  "inbound": {
```
mdebarros, line 6: Would UPPER-CASE to CAME-CASE be a breaking change here?

mdebarros (commented):
Hey @kleyow,

I don't see any major with the PR, just see my one comment about breaking changes.

config/development.json
kleyow: Unsure. So...I'm going to make it one.

kleyow:
> I also recommend some updates to documentation...I.e. perhaps add something explaining (and/or linking) these changes for PM4ML support.

I'll ask someone for the relevant documentation if there is any and add it to the readme.

src/server/start.ts
```diff
@@ -31,9 +31,30 @@ import { PACKAGE } from '../shared/config'
...
+  return server
+}
+
+export async function stop(server: Server): Promise<Server> {
+  await server.stop()
+  serverInstance = null
   return server
 }
+
+export async function restart(server: Server): Promise<void | Server> {
```
kirgene, line 50: I think it would be cleaner to have a factory class rather than separate methods in a module.
So instead of
```ts
import someModule from './server';
someModule.start(); // <--- it's unclear which module we access
```
we can have:
```ts
import { Server } from './server';
const server = Server::start(); // it's clear that we access Server class
```

kirgene (changes requested):

src/server/start.ts
kleyow: I'd rather just keep it functional for now vs extending off of `hapi.server`. I found it not that inituitive as well when messing about but lets leave that for a future refactor

src/reconfiguration/controlAgent/index.ts
```diff
@@ -0,0 +1,211 @@
...
+      msg = deserialize(data)
+    } catch (err) {
+      this.logger.push({ data }).log("Couldn't parse received message")
+      this.send(build.ERROR.NOTIFY.JSON_PARSE_ERROR(null))
+    }
+    this.logger.push({ msg }).log('Handling received message')
+    switch (msg.msg) {
+      case MESSAGE.CONFIGURATION:
+        switch (msg.verb) {
+          case VERB.NOTIFY:
```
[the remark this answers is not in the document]
kleyow: done.

package.json
```diff
@@ -38,11 +38,8 @@
     "release": "standard-version --releaseCommitMessageFormat 'chore(release): {{currentTag}} [skip ci]'",
     "snapshot": "standard-version --no-verify --skip.changelog --prerelease snapshot --releaseCommitMessageFormat 'chore(snapshot): {{currentTag}}'",
```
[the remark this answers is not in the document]
kleyow: Replaced with a fork that seems to have active contributions for now.

kirgene (approved):

